PR #44, opened 2018-04-14
Add csv exporting

Tvde1:
This is useful if you want to visualise data. Based on the branch @banCoder made (but it was outdated).


Tyrrrz:
I'm just looking at both PRs and wondering, wouldn't JSON output be much better than CSV?

Tvde1:
CSV is moe compact. JSON could also be an option I guess.

Tyrrrz:
Thanks for PR!

Issue #42, opened 2018-04-06
Doesn't export discord embeds


mdiller:
The tool currently doesn't support exporting discord embeds. It seems like this would be a bit of a challenge but doable. an example of something similar is this [Embed Visualizer](https://leovoel.github.io/embed-visualizer/)

Tyrrrz:
Related to #30

Tyrrrz:
Implemented by @mdiller in #46

PR #46, opened 2018-04-19
Added Embeds Support

mdiller:
Most of the work for converting the embeds to html was already done in this [Embed Visualizer](https://leovoel.github.io/embed-visualizer/), so once I converted that javascript into C#, I only had to make some adjustments to get it working nicely.

Let me know if there is anything I should do differently or anything.

One thing to note is that it currently doesn't support converting mentions in the embed descriptions. This is a bit tricky to add because we don't have the list of mentions like the main message does. We'll have to query to get users for that server I think. Let me know if you think of a better way to do it.

**Example Channel Conversion:**
I converted a channel from my bot's help server as an example:
Discord Channel Link: https://discord.gg/KAAfs2A
File (zipped because GitHub doesn't let me attach html): [Mangobyte Info - demos-examples.html.zip](https://github.com/Tyrrrz/DiscordChatExporter/files/1926885/Mangobyte.Info.-.demos-examples.html.zip)

**Example Embed:**

Embed output html:
image

Embed viewed in Discord App:
image

implements #42

### DIFF
--- a/DiscordChatExporter.Core/DiscordChatExporter.Core.csproj
+++ b/DiscordChatExporter.Core/DiscordChatExporter.Core.csproj
@@ -6,8 +6,13 @@
     </PropertyGroup>
 
     <ItemGroup>
+      <None Remove="Resources\ExportService\Shared.css" />
+    </ItemGroup>
+
+    <ItemGroup>
         <EmbeddedResource Include="Resources\ExportService\DarkTheme.css" />
         <EmbeddedResource Include="Resources\ExportService\LightTheme.css" />
+        <EmbeddedResource Include="Resources\ExportService\Shared.css" />
     </ItemGroup>
 
     <ItemGroup>
--- a/DiscordChatExporter.Core/Models/Embed.cs
+++ b/DiscordChatExporter.Core/Models/Embed.cs
@@ -34,18 +34,18 @@ namespace DiscordChatExporter.Core.Models
 
         public IReadOnlyList<EmbedField> Fields { get; }
 
-        public IReadOnlyList<User> MentionedUsers { get; }
+        public List<User> MentionedUsers { get; }
 
-        public IReadOnlyList<Role> MentionedRoles { get; }
+        public List<Role> MentionedRoles { get; }
 
-        public IReadOnlyList<Channel> MentionedChannels { get; }
+        public List<Channel> MentionedChannels { get; }
 
         public Embed(string title, string type, string description, 
             string url, DateTime? timeStamp, Color? color, 
             EmbedFooter footer, EmbedImage image, EmbedImage thumbnail, 
             EmbedVideo video, EmbedProvider provider, EmbedAuthor author, 
-            IReadOnlyList<EmbedField> fields, IReadOnlyList<User> mentionedUsers,
-            IReadOnlyList<Role> mentionedRoles, IReadOnlyList<Channel> mentionedChannels)
+            List<EmbedField> fields, List<User> mentionedUsers,
+            List<Role> mentionedRoles, List<Channel> mentionedChannels)
         {
             Title = title;
             Type = type;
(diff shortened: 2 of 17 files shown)


Tyrrrz:
Wow, thanks a lot for this. I'll make sure to take a closer look when I have more time today, but this looks amazing.

Tyrrrz:
Also, do you think we should also make some effort to export embeds into plaintext/csv?

mdiller:
No problem! I love your tool and was thinking it could be super useful for documentation for my discord bot if it had this feature. Didn't even take too long!

That sounds good to me, I'm just not sure what exactly we should put in them. Maybe some sort of combination of the text fields?

Tyrrrz:
Yeah, it's hard to represent embeds in plaintext while having them remain readable, but I guess we could somehow try concatenating all fields that have text in them.

mdiller:
@Tyrrrz hey by the way I was wondering. Why did you decide to not use one of the C# discord APIs ([DiscordSharp](https://github.com/suicvne/DiscordSharp) [DSharpPlus](https://github.com/DSharpPlus/DSharpPlus) [Discord.Net](https://github.com/RogueException/Discord.Net)) instead of building your own DataService and models?

Tyrrrz:
@mdiller I honestly don't remember my reasoning, but I remember checking out Discord.net long time ago and find it lacking somehow.

mdiller:
Makes sense. it's always nice working with stuff that you've made yourself so you know how it'll behave.

mdiller:
I moved the css that both LightTheme and DarkTheme have into the Shared.css file.

I also updated the mentions so they look more like discord mentions. I'm guessing we're trying to get things to look as much like discord as possible
image

mdiller:
I implemented most of what we need to get mentions working in embeds. I'm not sure how you want the last part of it implemented. The problem is we need to fetch the users/channels/roles that are mentioned in the embed, but I'm not sure where we should do that. Seems like there are 2 obvious options, each with downsides:
## Load all the users, roles, and channels that are in the guild into the caches before loading the messages.
The downside of this is that there are some servers with over 50k users, meaning we would have to do ~50 requests to cache all of them. It's likely that there are not many of them mentioned in embeds so this would be a bit of a waste.
## Retrieve the users, roles, and channels as we need them (inside the ParseEmbed function)
The downside of this is that it breaks up the format of the functions a bit, because ParseEmbed and ParseMessage would both have to be async methods and they would both have to be given an api token.

Which one do you think is better, or do you have another idea?

Tyrrrz:
Also thanks for tidying up and improving the CSS! To keep the PRs clean, please consider making separate PR requests if they are not directly related to this one (embeds) so that it will be easier to review/merge them. 😃

Tyrrrz:
Good point about Cli, I didn't even think of that. Hm.

I guess mentions can be tricky. Can you check how much slower would querying 50k users be? Worst case scenario we can leave the "mentioning users in embeds" feature out for now and come back to it later.

Tyrrrz:
That sounds reasonable, i think that's the way to go

Tyrrrz:
Sounds great :)

mdiller:
Done. It's not perfect, I had to make some adjustments because some endpoints aren't available for non-bots. Let me know what you think. I'll add something for csv and text exports as well. I'm thinking maybe we don't export to text because from what i can tell its mostly used for machine learning, and i'm guessing they wouldn't want embeds?

Tyrrrz:
Great, thanks! I'll check it out later this week.

Csv is used for machine learning for some other kind of parsing. PlainText is used when you want a very light-weight but readable export.

Tyrrrz:
Merged, thanks!
There's a lot of new code so I will make a bit of refactoring soon.

Tyrrrz:
Hi @mdiller 
I just got around to refactoring, can invite me to your server that you tested the embeds on? (Tyrrrz#5447)

mdiller:
sent!

Tyrrrz:
This feature is now released

PR #92, opened 2018-09-05
Dockerize the application

simnalamburt:
Thank you for making the awesome application! Since you have separated the CLI version, macOS and Linux users can use DiscordChatExporter conveniently by Docker!

```bash
docker run "-v${PWD}:/a" simnalamburt/discordchatexporter export -c <CHANNEL ID> -t <TOKEN>
```

###### Reference
- https://hub.docker.com/r/simnalamburt/discordchatexporter/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM mono:5
+WORKDIR /root/build
+COPY DiscordChatExporter.sln favicon.ico ./
+COPY DiscordChatExporter.Core DiscordChatExporter.Core
+COPY DiscordChatExporter.Cli DiscordChatExporter.Cli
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Awesome :)
It would be nice if you could integrate it with appveyor.yml so that it gets pushed to docker hub on build.

simnalamburt:
@Tyrrrz If you configure the [Docker Hub automated build](https://docs.docker.com/docker-hub/builds/), there's no need to change the neither appveyor script nor any other script. Give it a try!

Tyrrrz:
Cool! I'll look more into it later, because I don't have much time right now. :)

Tyrrrz:
https://hub.docker.com/r/tyrrrz/discordchatexporter/

PR #145, opened 2019-02-17
Implement a more sophisticated markdown parsing engine

### LINKED ISSUE #109: HTML export fails to format /shrug properly.
Messages containing the string `¯\_(ツ)_/¯` are incorrectly formatted as `&#175;\<i>(ツ&#175;` in the exported HTML, causing them to appear improper while also italicizing all text that follows it.

### LINKED ISSUE #110: Rendering of "_" is inaccurate
The HTML exporter does not properly replicate _'s behavior in Discord's markdown.
Test message:
```
test_test test_test
_test_
_t_est
t_e_st
test_test_
```
Discord's rendering: [view here](https://i.imgur.com/klMruGQ.png)
DiscordChatExporter HTML rendering:
```html
<div class="chatlog__content">
test<i>test test</i>test<br /><i>test</i><br /><i>t</i>est<br />t<i>e</i>st<br />test<i>test</i>

</div>
```

### LINKED ISSUE #113: HTML Export strips first line of code block even when it's not a language identifier
I exported a channel I was in using the HTML Export, and some of the text in codeblocks is missing.

Here's how it looks exported:
image

And here's how it looks in Discord:
image

Here it is in edit mode:
image

What I'm guessing is happening (I haven't looked at the code), is that whatever's on the same line as the triple backticks is interpreted as a language identifier for syntax highlighting, but according to Discord's [documentation](https://support.discordapp.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-), this should be included in the output:
image

### LINKED ISSUE #124: HTML: Wrong Link if it's in <>
If in a Discord Chat there is a hidden link like this:
`<https://www.google.com>`
Then the html will register `https://www.google.com>` as the link, instead of `https://www.google.com`

### LINKED ISSUE #130: Underline markup doesn't get rendered properly
The HTML export displays the message like this `__Underlined Text__` instead of actually underlining it.

### LINKED ISSUE #131: Strikedown markup doesn't get rendered properly.
Similar to #130 .
HTML renders it like this `~~strike down message~~` instead of ~~strike down message~~

### LINKED ISSUE #132: The markup-ignore character (\) gets ignored.
Markup still shows, despite having the markup-ignore character (\\) right in front of it.

### LINKED ISSUE #139: Spoilers don't get rendered
||spoiler|| is not rendered as a spoiler. It doesn't have to be that you need to click it, but at least change the background like a spoiler that is already clicked.

### LINKED ISSUE #142: Users are tagged in codeblocks
When you have a user mention inside of a codeblock it gets parsed to a real mention. It's not happening when writing `@username#1234` but with the id `<@!1234567890>` or `<@1234567890>`.

Tyrrrz:
The current, RegEx-based Markdown parser is suffering from a number of issues:
- It's quite hard to maintain
- It forces me to put a lot of render-specific code on the backend instead of templates
- It can't parse (sanely) certain constructs or combinations leading to a number of rendering inconsistencies

The idea is to implement a more sophisticated parsing engine that converts raw markdown into an AST-like form, which is then rendered separately by templates. This will make the code cleaner and also enable writing much more complex grammar rules.

Fixes #109
Fixes #110
Fixes #113
Fixes #124
Fixes #130
Fixes #131
Fixes #132
Fixes #139
Fixes #142
...and potentially others.


Tyrrrz:
It's funny, you would think that most of the time spent would be going towards making the parser work correctly, but instead, it's spent on making it work equally wrong as Discord's implementation.

I had to replace my node-based symmetrical recursive parsing approach with a regex-based non-greedy deferred recursing approach because of how Discord renders these corner cases:

- image
 is rendered as image (closing token is matched too early)
- image is rendered as image (backticks are not working properly)

PR #172, opened 2019-05-08
Add bot tags and some CSS fixes

mdiller:
- Fixes a couple things in the discord css
- Adds the discord font so everything looks nice
- Adds support for is_bot so that we can add bot tags in the html

Looks like this:
image

I'm not sure if we would want to add the is_bot to the other kinds of output, but I can add that if you want!

### DIFF
--- a/DiscordChatExporter.Core.Rendering/Resources/HtmlShared.css
+++ b/DiscordChatExporter.Core.Rendering/Resources/HtmlShared.css
@@ -341,3 +341,16 @@ img {
     margin-left: 6px;
     font-size: .875em;
 }
+
+.chatlog__bot-tag {
+    margin-left: 0.3em;
+    background: #7289da;
+    color: #ffffff;
+    font-size: 0.625em;
+    padding: 1px 2px;
+    border-radius: 3px;
+    vertical-align: middle;
+    line-height: 1.3;
+    position: relative;
+    top: -.2em;
+}
--- a/DiscordChatExporter.Core.Rendering/Resources/HtmlShared.css
+++ b/DiscordChatExporter.Core.Rendering/Resources/HtmlShared.css
@@ -289,7 +289,7 @@ img {
 }
 
 .chatlog__embed-footer-text {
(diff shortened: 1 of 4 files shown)


DiscordChatExporter.Core.Rendering/Resources/HtmlShared.css
```diff
@@ -1,5 +1,35 @@
 /* === GENERAL === */
 
+@font-face {
+    font-family: Whitney;
+    src: url(https://discordapp.com/assets/6c6374bad0b0b6d204d8d6dc4a18d820.woff);
```
Tyrrrz: Are you sure these URLs are static? Is there a way to embed them or would that make the report too big?

Tyrrrz:
Looks sweet @mdiller ❤️ 
I left one comment

DiscordChatExporter.Core.Rendering/Resources/HtmlShared.css
Tyrrrz: I'm unfortunately unfamiliar with `@font-face`, what would happen if the user doesn't have internet connection? Will it just fall back to the available fonts for this font family?

mdiller: Good question. I'm 100% sure that they're static, but I can't find a more reliable looking url to use. We could potentially host them on the repository's pages website thing? Idk. 

They're ~80KB each, and since a simple export for a channel is like 40KB, it's probably not worth it to embed these in the report. (80 * 5) = 400KB of extra data.

Good news is that yes `font-family` will just fall back to the next available font. In this case if it fails to get the Whitney, it will fall back to Arial like before. (I tested this to confirm by just typing stuff in the urls to make then invalid.)

Tyrrrz: Awesome, I think this is good enough then

Tyrrrz:
Thanks 🎉

PR #196, opened 2019-07-13
Add message ids in HTML export

### LINKED ISSUE #187: Message ids in html export
@Tyrrrz It would be useful if the messages dumped to html could include their message ids, so that i could programmatically find one given its id. Currently, messages arent even individually wrapped in a div, but it looks like it would be pretty trivial to add a div that wraps each message, and give that div an `id="message_1234567890"`. To be clear, this wouldn't look any different to someone just looking at the rendered html, and is just for if someone wants to be able to find a message programmatically.

I can implement this and create a pull request for it, I just wanted to check with you first to make sure it sounds good to you.

mdiller:
fixes #187

### DIFF
--- a/DiscordChatExporter.Core.Rendering/Resources/HtmlShared.html
+++ b/DiscordChatExporter.Core.Rendering/Resources/HtmlShared.html
@@ -75,6 +75,7 @@
 
                 {{~ # Messages ~}}
                 {{~ for message in group.Messages ~}}
+                    <div class="chatlog__message" id="message__{{ message.Id }}">
                     {{~ # Content ~}}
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Hi,
Can you indent the contents of the div you added?

Tyrrrz:
Also I'd prefer if the attribute was `data-message-id` instead of `id` to separate business logic from view logic

mdiller:
It looks like the indentation is based on the templating language, and that it's actually correct how it is I think?

Tyrrrz:
Yeah, that's a good point. But that means I was doing it wrong before. :)

PR #279, opened 2020-03-13
Add support for Guild Member object & included data

### LINKED ISSUE #153: Nicknames
Thank you for this! Is it possible to show members nicknames on my server in the downloaded file instead of the users main name? It's a bit difficult sometimes to keep track of who is who.

Thank you

ThereGoesMySanity:
The data included in the Guild Member object allowed me to implement colored names and nicknames in the HTML output.

I wasn't sure if JSON output should include either of these, so that's something that might still need to be done.

I had to refactor existing code in a couple important ways:

 - The only way to check if a user isn't a guild member without retrieving every guild member (which might require additional privileges) is to send the API request and check for a 404. Because anything other than 200 made the program throw an error, I had to add a "errorOnFail" bool to disable this functionality.

- I couldn't find a good place to save nickname and color since they depend on the guild and the user, so I had to add functions to the HTML message template to get these values.

Closes #153

### DIFF
--- a/DiscordChatExporter.Core.Models/Guild.cs
+++ b/DiscordChatExporter.Core.Models/Guild.cs
@@ -44,7 +44,7 @@ namespace DiscordChatExporter.Core.Models
                 )?.Where(r => r.Color != 0)?
                 .Aggregate(Role.Everyone, (a, b) => a.Position > b.Position? a : b)?
                 .ColorAsHex ?? "";
-        public static string GetUserNick(Guild guild, User user) => guild.Members[user.Id]?.Nick ?? user.Name;
+        public static string GetUserNick(Guild guild, User user) => guild.Members.GetValueOrDefault(user.Id)?.Nick ?? user.Name;
 
         public static string GetIconUrl(string id, string? iconHash)
         {
--- a/DiscordChatExporter.Core.Models/Guild.cs
+++ b/DiscordChatExporter.Core.Models/Guild.cs
@@ -42,7 +42,7 @@ namespace DiscordChatExporter.Core.Models
                             .Where(role => r == role.Id)
                             .FirstOrDefault()
                 )?.Where(r => r.Color != 0)?
-                .Aggregate(Role.Everyone, (a, b) => a.Position > b.Position? a : b)?
+                .Aggregate<Role, Role?>(null, (a, b) => (a?.Position ?? 0) > b.Position? a : b)?
                 .ColorAsHex ?? "";
         public static string GetUserNick(Guild guild, User user) => guild.Members.GetValueOrDefault(user.Id)?.Nick ?? user.Name;
 
--- a/DiscordChatExporter.Core.Models/Guild.cs
+++ b/DiscordChatExporter.Core.Models/Guild.cs
@@ -1,4 +1,8 @@
-﻿namespace DiscordChatExporter.Core.Models
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace DiscordChatExporter.Core.Models
 {
     // https://discordapp.string.IsNullOrWhiteSpace(com/developers/docs/resources/guild#guild-object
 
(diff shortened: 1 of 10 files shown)


ThereGoesMySanity:
New changes:
- Added coloring for role mentions

- When a role mention is colored, it becomes the same weight as normal message text.

| Before                                                                                                                                      | After                                                                                                                                       |
|---------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|
| Screenshot from 2020-03-13 18-42-07 | Screenshot from 2020-03-13 18-41-56 |
| Screenshot from 2020-03-13 18-41-08 | Screenshot from 2020-03-13 18-40-54 |

ThereGoesMySanity:
Fixes https://github.com/Tyrrrz/DiscordChatExporter/issues/153

Tyrrrz:
Hi. Finally found time to check this. Thank you for your contribution!

I see you're getting `Member` for each specific user by getting the list of guild members first. This is not optimal because the `Message` API object actually already contains `member` field, which we can use instead.

https://discordapp.com/developers/docs/resources/channel#message-object

Can you please update the implementation so that it reads the nickname and role IDs from the messages instead?

ThereGoesMySanity:
I tried this at first - the member object in messages is actually only provided if you're listening for message creation and message update events. When retrieving past messages, there's no member object.

Tyrrrz:
Ah I see. That sucks. :(

DiscordChatExporter.Core.Models/Guild.cs
```diff
@@ -10,13 +14,19 @@ public partial class Guild : IHasId
 
         public string? IconHash { get; }
 
+        public List<Role> Roles { get; }
```
Tyrrrz: Change this to `IReadOnlyList<Role>` for portability and consistency (will also have to update constructor).

DiscordChatExporter.Core.Models/Guild.cs
```diff
@@ -26,13 +36,23 @@ public Guild(string id, string name, string? iconHash)
...
+
         public static string GetIconUrl(string id, string? iconHash)
         {
             return !string.IsNullOrWhiteSpace(iconHash)
                 ? $"https://cdn.discordapp.com/icons/{id}/{iconHash}.png"
                 : "https://cdn.discordapp.com/embed/avatars/0.png";
         }
 
-        public static Guild DirectMessages { get; } = new Guild("@me", "Direct Messages", null);
+        public static Guild DirectMessages { get; } = new Guild("@me", "Direct Messages", new List<Role>(), null);
```
Tyrrrz: Change `new List<Role>()` to `Array.Empty<Role>` after updating the constructor (re previous comment).

DiscordChatExporter.Core.Services/ExportService.cs
```diff
@@ -50,8 +50,21 @@ public ExportService(SettingsService settingsService, DataService dataService)
...
+
+                foreach (User u in encounteredUsers)
+                {
+                    if(!guild.Members.ContainsKey(u.Id))
+                    {
+                        var member = await _dataService.GetGuildMemberAsync(token, guild.Id, u.Id);
+                        guild.Members[u.Id] = member;
+                    }
+                }
+
```
Tyrrrz: I'm thinking, maybe it will be easier at this point to just get all guild members? We can get them in batches of 1000 per request via API. That may be more efficient than making 1 request per each participant, and would probably make the code simpler too because we can remove `encounteredUsers` etc.

What do you think?

ThereGoesMySanity: I tried that as well - if you do that with a user token, Discord gives you a 403 forbidden and makes you reverify your email. I didn't test it with a bot token.

Tyrrrz: Ouch

Tyrrrz:
Thanks! 🎉

PR #287, opened 2020-04-07
Add option to use UTC instead of local time for timestamps

FM-96:
This adds an option in the settings menu to use UTC instead of the local time for message timestamps (defaults to false).

(This does not affect the JSON export format. However, since time zone information is preserved in that format, whoever consumes it can convert into their preferred time zone anyway.)


Tyrrrz:
You should be able to get UTC time as it is, by setting date format to `U`.

FM-96:
Welp, that's good to know. Guess this is unnecessary then. 😅

Tyrrrz:
No worries :)

PR #355, opened 2020-08-06
[GUI] Add controls to configure export time limits in addition to dates

### LINKED ISSUE #36: [GUI] More specific export time limits
First off I'd just like to thank you for the program, it's really nice, but I was wondering if it would be possible to have the option for a tighter export window? At the moment you can only limit it by day, would it be possible to add a day and time option?

CarJem:
image

Closes #36


Tyrrrz:
What happens currently if you select time but not the date?

CarJem:
It's Binding to the same date and time converter as the date pickers... so no crash... it will default to the current day.

Tyrrrz:
Is defaulting to the same day desirable? I think it would make more sense if the time pickers were disabled & had no effect if corresponding date pickers weren't used.

Tyrrrz:
I'll merge and apply some tweaks. Thanks!

CarJem:
No Problem!

Issue #363, opened 2020-08-19
Download multiple servers at once


lorenzyet:
I know there is the option to download all chats in a server, but how can I download all chats of all servers?

Tyrrrz:
![You can't do that](https://media1.tenor.com/images/6fe8ef8b3b4cd1853fff45826cf3cd23/tenor.gif?itemid=18020168)

wyattscarpenter:
@lorenzyet I made a bash script to do this, hope this helps! https://github.com/Tyrrrz/DiscordChatExporter/issues/274#issuecomment-638386963

PR #373, opened 2020-09-01
[CLI] Add command to export all channels across all guilds

### LINKED ISSUE #274: Export all channels from all guilds in a single action
Could you make a feature that allows people to export all channels, in servers, and dms, as a way to archive all chats locally?

wyattscarpenter:
Add a new command to export all Guilds. I don't really write C#, so I just combined the internal logic of GetGuildsCommand.cs and ExportGuildCommand.cs in a way that hopefully works! This addresses issues #274 and #363.

Presumably the syntax for calling this command will have to be added elsewhere, though I imagine it isn't hard.

Closes #274

### DIFF
--- a/DiscordChatExporter.Cli/Commands/ExportAllCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/ExportAllCommand.cs
@@ -3,6 +3,7 @@ using System.Threading.Tasks;
 using CliFx;
 using CliFx.Attributes;
 using DiscordChatExporter.Cli.Commands.Base;
+using DiscordChatExporter.Domain.Discord.Models;
 using DiscordChatExporter.Domain.Utilities;
 
 namespace DiscordChatExporter.Cli.Commands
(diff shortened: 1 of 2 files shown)


wyattscarpenter:
I couldn't really figure out if this will include DMs or not... I imagine people will want to export DMs as well, though perhaps in that case the file should be renamed `ExportAllCommand.cs`

Tyrrrz:
Thanks for the pr, @wyattscarpenter.
It looks like your code doesn't compile. You need to inherit from `ExportMultipleCommandBase` not `TokenCommandBase`.
I think there should be also a switch that dictates whether to include DMs or not (`--include-dm`) and the name of the command can be changed to `exportall`.

wyattscarpenter:
It inherits the right class now.

wyattscarpenter:
Now it should include the DMs by default, and has a flag to exclude them.

Tyrrrz:
Thanks!

PR #388, opened 2020-10-01
[GUI] Add dark theme

### LINKED ISSUE #298: [GUI] Add dark theme to the app
I searched and didn't see anyone requesting dark theme for the app itself so here I am, same as discord I can't stand light theme without some sunglasses so having an option in the app to switch to dark theme would really make me love this even more

andrewkolos:
Closes #298.

New toggle button for a Dark theme in the settings view:
image
The theme changes when toggled. The theme is a saved setting, and the saved theme is set when the app opens. Check out[ this album](https://imgur.com/a/eT4Qx11) to get a feel for what the dark theme looks like.

**Changes to light theme.** Due to the implementation I chose, there are some slight changes to the standard light theme. See [this album](https://imgur.com/a/peFNA6s) for images with captions that explain what's visually different.

**Implementation.** Given my zero prior experience with this software and WPF, I decided to reach for a simplest approach. I probably could have completely preserved the original colors by extracting the colors originally defined in App.xaml into a separate dictionary, defined another dictionary of colors for the dark theme, and then loaded/unloaded these at runtime to swap themes. Instead, I decided to remove all the custom colors/brushes and let the MaterialDesignInXaml (hereafter "MDIX") framework do the work of managing colors of text between dark and light modes.

First, I changed the installation of MDIX to more closely match the MDIX documentation. I chose a bundled theme of blue arbitrarily, which is going to get overwritten during startup by the root view model..
In the settings view, I added a new boolean member to the model and a new toggle button bound to that boolean. When the toggle button is clicked, the theme is changed by calling a method defined in `App.xaml.cs`. This method uses MDIX functionality to change the base theme (light/dark) and the primary color (the dark gray for light, and a light gray for dark).
RootViewModel now sets the theme in `OnViewLoaded`.

**Problems/concerns.**
* **Changes to the original theme's text coloring.** This implementation, as I mentioned, slightly changes how the standard light theme looks, since all the colors were deleted from App. This gives back MDIX control over text colors so that text is always readable regardless of whether Light or Dark is used as the base theme. Nonetheless, there is probably a way to preserve the original look completely by instead specifying separate color/brush dictionaries for both the light and dark theme and then loading/unloading them as needed. Given my inexperience with WPF, I reached for a simpler, less robust implementation to get this working. 
* **Theming from the code behind.** The themes aren't specified in xaml, but rather code (`Theme.cs`). The theme specified in `App.xaml` contains misleading dummy/placeholder colors. Maybe it would be better if both themes were written out clearly in the App.xaml markup (or their own files) to reduce coupling and indirection between the view's appearance and the code behind.
* Increased coupling. `RootViewModel.cs` and `SettingsView.xaml.cs` now both refer to `App.xaml.cs` directly for the theme changing functionality. While I'm not very experienced with WPF/MVVM, I have to imagine that the way I did this isn't the idiomatic way to accomplish theme changing.
Another smaller problem is that the theme gets redundantly re-updated when the settings view gets opened. I'm guessing its because the toggle button fires its Checked event at creation time when the value gets bound from the settings.

If/when these changes are acceptable. I'm happy to squash this to a single commit if preferred. Otherwise, feel free to test, tweak, and merge.

edit: typos

### DIFF
--- a/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
+++ b/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
@@ -109,7 +109,7 @@ namespace DiscordChatExporter.Gui.ViewModels
             }
             if (_settingsService.IsDarkThemeEnabled != null)
             {
-                ((App)System.Windows.Application.Current).setBaseTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
+                ((App)System.Windows.Application.Current).setTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
             }
 
             await HandleAutoUpdateAsync();
--- a/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
+++ b/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
@@ -107,10 +107,8 @@ namespace DiscordChatExporter.Gui.ViewModels
                 IsBotToken = _settingsService.LastToken.Type == AuthTokenType.Bot;
                 TokenValue = _settingsService.LastToken.Value;
             }
-            if (_settingsService.IsDarkThemeEnabled != null)
-            {
-                ((App)System.Windows.Application.Current).setTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
-            }
+            
+            ((App)System.Windows.Application.Current).setTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
 
             await HandleAutoUpdateAsync();
         }
--- a/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
+++ b/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
@@ -108,7 +108,7 @@ namespace DiscordChatExporter.Gui.ViewModels
                 TokenValue = _settingsService.LastToken.Value;
             }
             
-            Theme.SetAppTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
+            Theme.SetCurrent(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
 
             await HandleAutoUpdateAsync();
         }
--- a/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
+++ b/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
@@ -107,6 +107,10 @@ namespace DiscordChatExporter.Gui.ViewModels
                 IsBotToken = _settingsService.LastToken.Type == AuthTokenType.Bot;
                 TokenValue = _settingsService.LastToken.Value;
             }
+            if (_settingsService.IsDarkThemeEnabled != null)
+            {
+                ((App)System.Windows.Application.Current).setBaseTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
+            }
 
             await HandleAutoUpdateAsync();
         }
--- a/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
+++ b/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
@@ -108,7 +108,7 @@ namespace DiscordChatExporter.Gui.ViewModels
                 TokenValue = _settingsService.LastToken.Value;
             }
             
-            ((App)System.Windows.Application.Current).setTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
+            ((App)System.Windows.Application.Current).SetTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
 
             await HandleAutoUpdateAsync();
         }
--- a/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
+++ b/DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
@@ -108,7 +108,7 @@ namespace DiscordChatExporter.Gui.ViewModels
                 TokenValue = _settingsService.LastToken.Value;
             }
             
-            ((App)System.Windows.Application.Current).SetTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
+            Theme.SetAppTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
 
             await HandleAutoUpdateAsync();
         }
--- a/DiscordChatExporter.Gui/App.xaml
+++ b/DiscordChatExporter.Gui/App.xaml
@@ -21,7 +21,6 @@
                 <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />
             </ResourceDictionary.MergedDictionaries>
 
-
             <!--  Styles  -->
             <Style x:Key="MaterialDesignRoot" TargetType="{x:Type Control}">
                 <Setter Property="FontFamily" Value="{DynamicResource MaterialDesignFont}" />
--- a/DiscordChatExporter.Gui/App.xaml
+++ b/DiscordChatExporter.Gui/App.xaml
@@ -15,9 +15,9 @@
             <!--  Merged dictionaries  -->
             <ResourceDictionary.MergedDictionaries>
                 <materialDesign:BundledTheme
-                    BaseTheme="Dark"
-                    PrimaryColor="LightBlue"
-                    SecondaryColor="LightBlue" />
+                    BaseTheme="Light"
+                    PrimaryColor="Blue"
+                    SecondaryColor="Blue" />
                 <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />
             </ResourceDictionary.MergedDictionaries>
 
--- a/DiscordChatExporter.Gui/App.xaml
+++ b/DiscordChatExporter.Gui/App.xaml
@@ -14,68 +14,13 @@
 
             <!--  Merged dictionaries  -->
             <ResourceDictionary.MergedDictionaries>
-                <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Light.xaml" />
+                <materialDesign:BundledTheme
+                    BaseTheme="Dark"
+                    PrimaryColor="LightBlue"
+                    SecondaryColor="LightBlue" />
                 <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />
             </ResourceDictionary.MergedDictionaries>
 
-            <!--  Colors  -->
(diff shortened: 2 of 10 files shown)


Tyrrrz:
Thanks @andrewkolos! I will review this in a bit.

Tyrrrz:
First of all, thanks for such a nice and detailed description of the changes and the PR itself!

Regarding changes to the old theme:

1. The lost contrast between fonts in the export dialog concerns me.

image
image

Is there anything that can be done to make it look like it did previously?

Also on the main screen, the contrast between different font styles is really nice but it seems to be completely lost in the light theme.

image

2. I originally also went with the hamburger button, but decided not to use it because of the back button. I guess it's not that big of a deal though.

3. I don't really like how the area around the input textbox looks in the dark theme. Maybe we can also make it dark, but in a bit different shade so it contrasts with surroundings? Currently it looks a bit like the colors were inverted.

image

DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
```diff
@@ -107,6 +107,8 @@ protected override async void OnViewLoaded()
                 IsBotToken = _settingsService.LastToken.Type == AuthTokenType.Bot;
                 TokenValue = _settingsService.LastToken.Value;
             }
+            
+            ((App)System.Windows.Application.Current).SetTheme(_settingsService.IsDarkThemeEnabled ? Theme.Dark : Theme.Light);
```
Tyrrrz, line 111: I would abstract this away to a static method in our `Theme` class, i.e. `Theme.SetCurrent(...)`.

Tyrrrz:
Overall implementation looks good to me!

DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
andrewkolos: Yeah--I think that's a good idea

andrewkolos:
Thanks for the feedback. I agree that text contrast is an issue. If you think it's not so good on the light theme exports dialog, then don't look at the dark theme version of it--it's the opposite of what we want!

The challenge here is that, previously, color contrast in text and controls was achieved by using the primary color. Take the Token textbox at the top for example. 
> Currently it looks a bit like the colors were inverted

Turns out this is exactly what I did. My reason was that, with how coloring currently works, the theme's primary color needs to contrast well with the background color. If we set the primary color to a dark gray while using the dark theme, then a lot of UI elements (e.g. buttons) wouldn't be easily seen against the dark background.
image
This problem didn't happen with the original light theme, because the dark gray chosen as the primary color contrasted well with the white background. This problem is the same reason I had to have text stop using the primary color as well. 
As a result, unless we want to choose a different primary color altogether, we need to keep the primary color for the dark theme as a lighter gray.

I'm sure there is a way to fight MDIX and get exactly what we want, but it would require a heavier, more powerful solution.
* **Theme switching mechanism**. We would need to do more than just change the primary/secondary colors of the MDIX theme. We would have to define unique colors for both the light theme and dark theme (e.g. `AccentTextForeground`, `TokenBoxForeground`, `SettingsCogWheelForeground`, etc.). All this information would take the shape of `ResourceDictionary`s defined in the markup and/or code. We would add/remove the light/dark theme dictionaries as needed to get the appearance to change. This is reminiscent of the alternate implementation I discussed in my original post.
* **Accentuating text.** While implementing a more powerful theme-switching mechanism would solve this problem. I do want to consider a workaround. Instead of relying on color to accentuate text, we could use different `FontWeight`s (e.g. `Light` for de-emphasized text, and `Bold` for emphasized text). But, if we got custom-theming for the token textbox to be desirable, then getting colors to work in a similar manner probably wouldn't be much more work at that point.

I'll tinker around and see if I can get this working :)

andrewkolos:
Here is an example of using font weights instead of using different text colors.
image

andrewkolos:
On second thought, I'll wait for your opinion before I start working on this.

Tyrrrz:
Font weights seem to work around the issue well! Can also go the opposite way and use lighter than normal fontweight on some elements to see if it fits well.

If I understood correctly, the problem with the dark theme is the primary color we used in the light theme? If that's the case, I'm fine with dark theme using its own primary color, they don't have to match exactly.

From the last two screenshots, it seems that fontweight could solve the contrast issue just fine.

andrewkolos:
> If I understood correctly, the problem with the dark theme is the primary color we used in the light theme? If that's the case, I'm fine with dark theme using its own primary color, they don't have to match exactly.

Yeah, pretty much. The primary/accent colors needs to contrast well with the background. The dark primary color of the light theme contrasts well with the light background. For a dark theme, we want a light (or at least neutral) primary color to contrast well with the dark background.

I completely understand the potential for confusion, because there are so many ways to approach theming. I'm going to re-summarize for the sake of clarity, so forgive me if I'm saying a bunch of things you already know.

We have two issues that need to be addressed here:
1. To maintain controls being able to contrast well against the background, the primary/secondary colors need to contrast well with the background color (light/dark).
2. To use (the primary) color to accentuate text, we need to pick a color that, again, contrasts well against the background. There is an additional challenge here: currently, we choose a darker shade of the primary color (`PrimaryHueDarkBrush`) to give a text a "bold" look. However, for dark mode, we actually want to use a lighter hue of the primary color, not a darker one.

**Ways to address problem 1.** 
1. Keep what I did initially: simply invert the primary color, resulting in that light gray. From here, we can either
    1. accept the bright-looking Token textbox and move on, or
    2. put in some extra work to custom-theme the Token textbox so that it stays dark while keeping the primary color as that light gray I chose.
2. Something I have yet to propose: Instead of having different primary colors for the dark and light themes respectively, we choose a new primary color that contrasts well in either mode. Take Discord itself as an example, they use this mid-light purple that looks OK in both modes.
image
image
If we did this, it would dramatically change how the app looks.

**Ways to address problem 2.** Regardless of the way we approach problem 1, problem 2 is going to exist as I wrote it above. Here's how we could approach it:
1. Keep it simple. As I said in my previous post, we could stop using color to achieve emphasis and just use font weight. One could argue that font weight is the more natural way to achieve our goal.
2. Get heavy-handed. In an analogous manner to how we could custom-theme the token textbox, we specify a dictionary for each theme. If we ended up going with the extra-theming approach to address problem 1, this isn't much more work and wouldn't introduce much more complexity--we just need to add brushes for text. We could also calculate these at runtime by using `MaterialDesignBody` as a base. We use this instead of primary color to remove the coupling between the emphasis effect in the primary color. Why is this potentially useful? Consider this example based on the current state of the app: say we change our primary color to a light purple, then the text we want to be emphasized doesn't look stronger--just different.
image
If you never planned to change the color of the app, then this benefit doesn't mean much but is something to consider.

**Choosing a way forward.** We could mix-and-match the solutions to address both problems to create a overall solution, but if that is overwhelming (it certainly took me a lot of time to sort out in my head)--I propose two choices that fall on the ends of the spectrum of simplicity versus capability:
1.  Simplicity. Accept the Token textbox being a light-gray color in the dark theme as I currently have it, and use font weight to (de)emphasize text rather than color. I would just need to make one more simple commit to change the text.
2. Capability. We add mechanism that takes responsibility for coloring the token textbox as well as the emphasized text. This could take the form of separate resource dictionaries that we swap out at runtime.

Thoughts? I'm OK pursuing any route, but I would obviously need a lot more time to implement something more custom/powerful.

Tyrrrz:
Again, thanks a lot for such detailed overview :)

> To maintain controls being able to contrast well against the background, the primary/secondary colors need to contrast well with the background color (light/dark).

Makes sense!

> To use (the primary) color to accentuate text, we need to pick a color that, again, contrasts well against the background. There is an additional challenge here: currently, we choose a darker shade of the primary color (PrimaryHueDarkBrush) to give a text a "bold" look. However, for dark mode, we actually want to use a lighter hue of the primary color, not a darker one.

Makes sense too.

> Keep what I did initially: simply invert the primary color, resulting in that light gray. From here, we can either
accept the bright-looking Token textbox and move on, or
put in some extra work to custom-theme the Token textbox so that it stays dark while keeping the primary color as that light gray I chose.

I really dislike the way the textbox looks currently in dark theme :/

I guess I've already asked this, but could we change the primary/secondary colors ONLY for Dark theme to make the colors fit better together?

> Something I have yet to propose: Instead of having different primary colors for the dark and light themes respectively, we choose a new primary color that contrasts well in either mode. Take Discord itself as an example, they use this mid-light purple that looks OK in both modes.

It looks more like an accent color, although they do use it in a lot of places.

> Keep it simple. As I said in my previous post, we could stop using color to achieve emphasis and just use font weight. One could argue that font weight is the more natural way to achieve our goal.

I'm fine with this. In your screenshots font weight seems to achieve the goal just fine.

> Get heavy-handed. In an analogous manner to how we could custom-theme the token textbox, we specify a dictionary for each theme that specifies text color. If we ended up going with the extra-theming approach to address problem 1, this isn't much more work and wouldn't introduce much more complexity--we just need to add brushes for text. We could also calculate these at runtime by using MaterialDesignBody as a base. We use this instead of primary color to remove the coupling between the text emphasis effect and our choice for the primary color. Why is this potentially useful? Consider this example based on the current state of the app: say we change our primary color to a light purple, then the text we want to be emphasized doesn't look stronger--just different.

I like the other solution more.

andrewkolos:
> I guess I've already asked this, but could we change the primary/secondary colors ONLY for Dark theme to make the colors fit better together?

Totally, that's an option I left out (I wasn't lying when I said I had a hard time juggling all the options in my mind :)). In the most recent commit, I chose another primary color for the dark theme. I chose blue for no good reason, so feel free to experiment with colors by messing around with the values for `Dark` in `Theme.cs` until you find something you like. I also implemented the font weight change. Feel free to test and tinker. Let me know if I missed anything!

Tyrrrz:
Ok I've spent a few days playing with this and here are my observations:
- MaterialDesignInXAML sucks.

The main issue is that, when left to its own devices, it sets the text color based on primary hue. Because of that, we're hitting the problem where the text color is tied to the color of other unrelated elements.

If you look at Discord, for example, notice that their brand color (shade of blue) remains the same in both themes, but it DOESN'T affect the text color:

image
image

It looks like around 4 years ago I had the exact same idea, which is why I redefined every color manually and made sure that primary hue setting doesn't affect the text color:

https://github.com/Tyrrrz/LightBulb/blob/fed2f898257b8af98dbfe39a41ae1c7bf174a2cc/LightBulb/App.xaml#L21-L77

Of course, with two themes this is no longer a suitable solution.

From here on I basically see two options:

- Implement a completely custom version of `PalleteHelper.SetTheme()` that doesn't infer text color from primary hue. Basically do this manually: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/c2abdfbdbab3397e0ec626d7272ee7edab5950cd/MaterialDesignThemes.Wpf/ResourceDictionaryExtensions.cs#L15-L77

- Drop dark theme support as it's already too much effort than it's probably worth. I know we've already committed some time to this (mostly you) but it's annoying how hard this is :D

Tyrrrz:
Although it looks like the problem really exists with the `MaterialDesignFlatButton`, which just uses `PrimaryHueMidBrush` for foreground. Other text appears fine.

image

Tyrrrz:
Nevermind I think I got it.

Tyrrrz:
Dark:
image
image

Light:
image
image

Tyrrrz:
Note that the settings button is actually of a bit different color than the others. But I think it's close enough to not be noticeable.

Tyrrrz:
Note: I still hate MDIX.

Tyrrrz:
Thanks for your work!

PR #403, opened 2020-10-10
[JSON] Add embed color

### LINKED ISSUE #396: [JSON] Embeds missing color
The information provided in JSON exports omits the color of embeds.  I know that the exporter does read this information since the HTML exports correctly include the color stripe on the left side of the embed.  Is there any way we could get the embed color information included in the JSON?

Thanks!

andrewkolos:
Closes #396.

Adds `color` field to `embeds` items in JSON output. Color is stored as an HTML hex color code. The `color` field will be omitted if the embed doesn't have a color.

image


Tyrrrz:
Awesome, thanks!

PR #405, opened 2020-10-10
Add "unix" and "unixms" options for date format

### LINKED ISSUE #220: Unix timestamp support
It would be amazing to have the option to use unix timestamps. This would make it much easier to migrate channel data to slack via csv import.

andrewkolos:
Closes #220.
For the `--dateFormat` option, adds values:
* `unix` for a unix timestamp (seconds since epoch).
* `unixms` for milliseconds since epoch.

Example using `--dateformat unix`:
image


Tyrrrz:
Thanks!

PR #410, opened 2020-10-14
Support multiple export formats

### LINKED ISSUE #265: Allow multiple output formats
Hey, I really like the html version for the ease of use, but the json version for better processing.

As scraping the discord might be a bad thing I'd rather avoid running it twice for everything.

if I specify `-f json -f html` I however get the error:
```
Target type is not enumerable and can't accept more than one value.
```

Other possible solution would be to have a command making a html file from the json in an offline fashion.

andrewkolos:
Closes #265.

Allows multiple export formats (only via CLI).

When multiple values are provided for `-f`, the value for `-o` will always be interpreted as a path to a directory. The filename of each exported file cannot be specified in this case.


Tyrrrz:
Hi! Looks like there are some conflicts with master.

Tyrrrz:
I'm looking at this and thinking: maybe a better approach is to internally create multiple `ExportRequest` instances (one for each format) and allow `MediaExporter` to take an array of those?

Tyrrrz:
Ah `ExportRequest` contains `ChannelId` so we probably need some higher-level structure to represent a group of requests bound to a single channel.

andrewkolos:
To me, an `ExportRequest` is first and foremost is a representation of the user input. If the user requests the data to be exported to multiple formats, than `ExportRequest` should contain all of the formats. If downstream code is only interested in one of the formats (or, more generally, some strict subset of information within the request), then, yes, I agree there is a missing abstraction(s) between `ExportRequest` and that code.

I feel like this is something I meant to bring up in the original PR, but I think I got frustrated trying to document all the uses and come up with a complete refactoring within the amount of time I had set aside for it. I will try to take another look at this and come up with a refactoring when I can find the time.

Tyrrrz:
You are right, `ExportRequest` is a poor place to plug this in. What I'm thinking is to create an abstraction that allows us to treat these multiple exports essentially as multiple requests, but reusing the same export context (including message stream, etc.). I want to avoid having the concept of multiple formats deeper in the domain, if that makes sense.

andrewkolos:
Yeah--I completely agree.

Tyrrrz:
Decided to descope it for next release (which is out btw, with all the stuff you've contributed!). 🎉 

I think we should really consider this feature more critically from the standpoint of introducing unnecessary complexity. After all, exporting in multiple formats is probably a rather niche use case, so it would be unfortunate if we had to introduce so too much complexity in exchange for it.

I'm still leaning towards this idea of an abstraction on top of `ExportRequest` that `ChannelExporter` would somehow consume to reuse the same message stream for multiple `MessageWriter` instances. I really don't want an array of formats to get deep into the domain, because then it will be annoying to deal with (as you've probably noticed with this PR).

I suggest dropping this PR for now, at least because I've introduced significant merge conflicts with the last refactoring. 😄

andrewkolos:
Closing for now. May or may not come back to this later.

PR #411, opened 2020-10-16
[GUI] Fix content bg color on first screen to match the second screen

andrewkolos:
Before:
image

After:
image

Apologies for the blue-light filtering in the images :).


Tyrrrz:
Thanks!

PR #416, opened 2020-10-25
[JSON] Add core user mentions to output

### LINKED ISSUE #415: Store mention ID in JSON
Hello! First of all, thanks for this great piece of software.

I am doing some parsing of messages on a server where I need to be able to link mentions to their original user in a permanent way. The way DiscordChatExporter currently works, mentions only get exported as "@nickname", which is prone to change very quickly. Would it be possible to get the full user ID, at least in the JSON format? Since I assume most people using JSON are looking for a way to process the data later, I feel it would also make sense (if the Discord API exposes this data).

For the implementation, I would suggest having an additional field `mentions` where all mentions present in the message/embed are dumped in sequential order. Something like this for example:

```json
{
  ...
  "content": "Hello @User1! Welcome to this server owned by @User2",
  "mentions": [
    {
      "id": 01234567890123456,
      "name": "User1",
      "discriminator": "1234",
      "nickname": "New User"
    },
    {
      "id": 01234567890123457,
      "name": "User2",
      "discriminator": "5678",
      "nickname": "The Owner"
    },
  ],
  ...
}
```

andrewkolos:
Partially addresses #415.

image

image

### DIFF
--- a/DiscordChatExporter.Domain/Exporting/Writers/JsonMessageWriter.cs
+++ b/DiscordChatExporter.Domain/Exporting/Writers/JsonMessageWriter.cs
@@ -4,6 +4,7 @@ using System.Threading.Tasks;
 using DiscordChatExporter.Domain.Discord.Models;
 using DiscordChatExporter.Domain.Exporting.Writers.MarkdownVisitors;
 using DiscordChatExporter.Domain.Internal.Extensions;
+using DiscordChatExporter.Domain.Markdown.Ast;
 
 namespace DiscordChatExporter.Domain.Exporting.Writers
 {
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Hi.

Thanks for another PR!

These mentions unfortunately don't include the ones in embeds because those aren't structured. In fact, DCE resolves mentions in embeds in a very awkward way, where it collects a list of users it has seen thus far and then parses the content and tries to map any mentions to that list.

Here's that part: https://github.com/Tyrrrz/DiscordChatExporter/blob/948fab7c8644bcdcc13936665413f72d64e0ec39/DiscordChatExporter.Domain/Exporting/ChannelExporter.cs#L42-L53

And here's how the mentions are parsed in the first place: https://github.com/Tyrrrz/DiscordChatExporter/blob/948fab7c8644bcdcc13936665413f72d64e0ec39/DiscordChatExporter.Domain/Exporting/Writers/MarkdownVisitors/PlainTextMarkdownVisitor.cs#L69-L77

(the content is parsed into AST and then a visitor goes through the nodes to locate and transform the mentions with the list of users from the context)

This PR is still useful as it covers everything but the embeds, but it doesn't fully cover the linked issue. I'll merge it regardless! Thanks again!

PR #432, opened 2020-11-13
[HTML] Rich rendering support for audio and video embeds

### LINKED ISSUE #333: [HTML] Rich attachment previews
Tried the exporter w/ media again and I noticed how ugly the current system looks for file attachments that embed like music files or mp4/webm and thought it would be nice to have them appear as they do in Discord so anyone viewing the HTML can click on the sound/video and it will play without having the user download it.

image

Also while I'm here talking about attachments, it would be nice if the downloader kept all of the original file names and stored them in the proper folders such as Demez's archiver does except you could probably organize things even better into file type like JPG, PNG, MP4, MP3, etc.

image

When a file is the same name as another previously downloaded file it adds a (x) to the front so if there's 50 files named "Unknown" then it will name them "Unknown (31)" etc, this is much better than all of the downloads being random numbers and letters.

sas41:
Closes #333

The changes make it so that the HTML writer embeds some web safe audio and video files.

In addition to that the ".webp" file format is also added to image embeds, which is on par with the default behaviour of Discord.

The new supported formats are:

Image:

- .WEBP

Audio:

-  .MP3
- .WAV
- .OGG
- .FLAC
- .M4A

Video:

- .MP4
- .WEBM

Image Previews:

Audio:
image

Video:
image

The file name and size are still there (above the embed) to help with searching and to keep it consistent, across all types of embeds. The same applies for images too:

Image:
image

### DIFF
--- a/DiscordChatExporter.Domain/Discord/Models/Attachment.cs
+++ b/DiscordChatExporter.Domain/Discord/Models/Attachment.cs
@@ -23,6 +23,11 @@ namespace DiscordChatExporter.Domain.Discord.Models
         public bool IsImage =>
             ImageFileExtensions.Contains(Path.GetExtension(FileName), StringComparer.OrdinalIgnoreCase);
 
+        public bool IsVideo =>
+            WebSafeVideoFileExtensions.Contains(Path.GetExtension(FileName), StringComparer.OrdinalIgnoreCase);
+        public bool IsAudio =>
+            WebSafeAudioFileExtensions.Contains(Path.GetExtension(FileName), StringComparer.OrdinalIgnoreCase);
+
         public bool IsSpoiler =>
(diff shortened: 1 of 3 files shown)


Tyrrrz:
Nice! Does it also handle spoilers well?

sas41:
> 
> 
> Nice, thank you! Does it also handle spoilers well?

I made some further changes to fix that, as I had missed it when I made the PR.

However, I had to make some changes to get the spoilered attachments to display correctly and consistently in Firefox.
The changes would not have made a difference if it hadn't been for a bug (or perhaps intended behaviour) in Firefox.
The fix I mustered had some aesthetic changes across the board, I _personally_ think that it's fine, but hey, it's your project, you decide.

Here is how the elements look now:
image

image

Cheers and thanks for the great OSS.

Tyrrrz:
Thank you!

PR #433, opened 2020-11-13
[CLI] Print error and exit if reuse-media option is used without media option

### LINKED ISSUE #425: [CLI] Report an error if `--reuse-media` is used without `--media`
Related to #424

Suggested approach: override the default `ExecuteAsync()` in `ExportCommandBase` (https://github.com/Tyrrrz/DiscordChatExporter/blob/547bf0a463c8f210d1cb943fb41fc00fe5a1facd/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs#L12) and add this extra validation.

Make sure all derived commands call `base.ExecuteAsync()`.

andrewkolos:
Closes #425.

`export -c <channel id> --token <token> --reuse-media` results in:
image

### DIFF
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -75,8 +75,7 @@ namespace DiscordChatExporter.Cli.Commands.Base
         {
             if (ShouldReuseMedia && !ShouldDownloadMedia)
             {
-                Console.WriteLine("The \"--reuse-media\" option cannot be used without the \"--media\" option.");
-                Environment.Exit(1);
+                throw new CommandException("The --reuse-media option cannot be used without the --media option.");
             }
         }
         protected async ValueTask ExportAsync(IConsole console, Channel channel)
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -69,6 +69,34 @@ namespace DiscordChatExporter.Cli.Commands.Base
             console.Output.WriteLine("Done.");
         }
 
+        public void ExecuteAsync()
+        {
+            if (ShouldReuseMedia && !ShouldDownloadMedia)
+            {
+                var shouldDownloadMediaOptionName = GetAttributeName(nameof(ShouldDownloadMedia));
+                var shouldReuseMediaOptionName = GetAttributeName(nameof(ShouldReuseMedia));
+
+                Console.WriteLine($"The \"--{shouldReuseMediaOptionName}\" option cannot be used without the \"--{shouldDownloadMediaOptionName}\" option.");
+                Environment.Exit(1);
+            }
+
+            string GetAttributeName(string name)
+            {
+                var property = typeof(ExportCommandBase).GetProperty(name);
+                if (property == null) throw new Exception("Could not find property: " + name);
+
+                var attributes = property.GetCustomAttributes(true);
+                foreach (var attribute in attributes)
+                {
+                    var commandOptionAttribute = attribute as CommandOptionAttribute;
+                    if (commandOptionAttribute != null && commandOptionAttribute.Name != null)
+                    {
+                        return commandOptionAttribute.Name;
+                    }
+                }
+                throw new Exception("Unable to find name of property");
+            }
+        }
         protected async ValueTask ExportAsync(IConsole console, Channel channel)
         {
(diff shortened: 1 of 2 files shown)


DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -69,6 +69,34 @@ protected async ValueTask ExportAsync(IConsole console, Guild guild, Channel cha
...
+                foreach (var attribute in attributes)
+                {
+                    var commandOptionAttribute = attribute as CommandOptionAttribute;
+                    if (commandOptionAttribute != null && commandOptionAttribute.Name != null)
+                    {
+                        return commandOptionAttribute.Name;
+                    }
+                }
+                throw new Exception("Unable to find name of property");
+            }
```
Tyrrrz, line 98: It's fine to use hardcoded values. If we really wanted to, we could use `const string MediaOptionName` and then use it in both places, but I don't think it's necessary as option names are not going to change since it's a breaking change.

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -69,6 +69,34 @@ protected async ValueTask ExportAsync(IConsole console, Guild guild, Channel cha
...
 
+        public void ExecuteAsync()
+        {
+            if (ShouldReuseMedia && !ShouldDownloadMedia)
+            {
+                var shouldDownloadMediaOptionName = GetAttributeName(nameof(ShouldDownloadMedia));
+                var shouldReuseMediaOptionName = GetAttributeName(nameof(ShouldReuseMedia));
+
+                Console.WriteLine($"The \"--{shouldReuseMediaOptionName}\" option cannot be used without the \"--{shouldDownloadMediaOptionName}\" option.");
+                Environment.Exit(1);
```
Tyrrrz, line 80: This can be replaced with:

```suggestion
                throw new CommandException($"The --reuse-media option cannot be used without the --media option.");
```

DiscordChatExporter.Cli/Commands/ExportChannelCommand.cs
```diff
@@ -12,7 +12,10 @@ public class ExportChannelCommand : ExportCommandBase
             Description = "Channel ID.")]
         public string ChannelId { get; set; } = "";
 
-        public override async ValueTask ExecuteAsync(IConsole console) =>
+        public override async ValueTask ExecuteAsync(IConsole console)
+        {
+            base.ExecuteAsync();
```
Tyrrrz, line 17: Missing `await` here.

Tyrrrz (changes requested):

DiscordChatExporter.Cli/Commands/ExportChannelCommand.cs
Tyrrrz: Also, is this the only place which needs to be updated? Are there any others where we need to call `base.ExecuteAsync()`?

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
andrewkolos: [I think attribute arguments have to be constant expressions.](https://stackoverflow.com/questions/16710533/passing-static-array-in-attribute), so we can't have a `MediaOptionName` variable and pass it into the `CommandOption`. We could just use the hard-coded `"reuse-media"` twice in the code, though.

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
andrewkolos: Oh, I thought throwing an exception would result in an ugly console message, but apparently it doesn't. Neat!

andrewkolos:
Ready for review again.

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
Tyrrrz: It was made specifically for this :)

Tyrrrz:
@andrewkolos still missing an `await` on `base.ExecuteAsync()`, otherwise looks great!

andrewkolos:
Hey, sorry for taking so long to get back to this. `ExportCommandBase.ExecuteAsync` is synchronous, so it cannot be `await`ed. We could mark it as async regardless and await on it, but Visual Studio would give a warning for marking synchronous code as asynchronous.

Tyrrrz:
Ah, the naming confused me. I'll change it to something else. Thanks!

PR #448, opened 2020-12-06
[HTML] Prettier attachment preview for non-media files

### LINKED ISSUE #436: [HTML] Prettier non-interactive attachment preview
Currently, the attachments which are neither images, video or audio, are previewed like this:

image

This looks a bit ugly. For comparison, this is how it looks on Discord:

image

We could try to make it prettier. No need to make it look exactly like the Discord version, but something a bit more presentable would be nice.

96-LB:
Closes #436 

Updates the preview for non-interactive files which were previously displayed as a link. The new appearance is the following:

image
image

For the sake of comparison, here's how the previews look on Discord:

image
image

The simple SVG icon is embedded into the page rather than being hosted remotely. Also, this pull request modified FileSize to display "bytes" in place of "B" to be consistent with Discord.

### DIFF
--- a/DiscordChatExporter.Domain/Discord/Models/Common/FileSize.cs
+++ b/DiscordChatExporter.Domain/Discord/Models/Common/FileSize.cs
@@ -52,7 +52,7 @@ namespace DiscordChatExporter.Domain.Discord.Models.Common
             if (Math.Abs(TotalKiloBytes) >= 1)
                 return "KB";
 
-            return "B";
+            return "bytes";
         }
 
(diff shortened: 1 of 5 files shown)


Tyrrrz:
Hi. Thanks for the PR!
Can we somehow extract the SVG code out (maybe inline it in CSS?) so it doesn't get repeated for every attachment?

96-LB:
I pulled out the path data/styling into the core stylesheet and condensed the SVG code. I've admittedly done little with embedded SVG in the past, so if there's a better way to do this, I'm all ears.

Tyrrrz:
Thanks!

PR #455, opened 2020-12-14
Add support for replies

### LINKED ISSUE #440: Add support for replies
The exporter works just fine with the old style of "quoting" messages, but it's not able to recognize the new reply feature.

Sanqui:
- [x] Parse message references
- [x] Output message references in JSON
- [x] Understand "reply" message type
- [x] Parse referenced messages
- [x] Output referenced messages in HTML

Closes #440

### DIFF
--- a/DiscordChatExporter.Domain/Discord/Models/Message.cs
+++ b/DiscordChatExporter.Domain/Discord/Models/Message.cs
@@ -18,7 +18,8 @@ namespace DiscordChatExporter.Domain.Discord.Models
         ChannelNameChange,
         ChannelIconChange,
         ChannelPinnedMessage,
-        GuildMemberJoin
+        GuildMemberJoin,
+        Reply = 19
     }
 
     // https://discord.com/developers/docs/resources/channel#message-object
@@ -50,6 +51,8 @@ namespace DiscordChatExporter.Domain.Discord.Models
 
         public MessageReference? Reference {get; }
 
+        public Message? ReferencedMessage {get; }
+
         public Message(
             string id,
             MessageType type,
@@ -63,7 +66,8 @@ namespace DiscordChatExporter.Domain.Discord.Models
             IReadOnlyList<Embed> embeds,
             IReadOnlyList<Reaction> reactions,
             IReadOnlyList<User> mentionedUsers,
-            MessageReference? messageReference)
+            MessageReference? messageReference,
+            Message? referencedMessage)
         {
             Id = id;
             Type = type;
@@ -78,6 +82,7 @@ namespace DiscordChatExporter.Domain.Discord.Models
             Reactions = reactions;
             MentionedUsers = mentionedUsers;
             Reference = messageReference;
+            ReferencedMessage = referencedMessage;
         }
 
         public override string ToString() => Content;
@@ -95,6 +100,7 @@ namespace DiscordChatExporter.Domain.Discord.Models
             var type = (MessageType) json.GetProperty("type").GetInt32();
             var isPinned = json.GetPropertyOrNull("pinned")?.GetBoolean() ?? false;
             var messageReference = json.GetPropertyOrNull("message_reference")?.Pipe(MessageReference.Parse);
+            var referencedMessage = json.GetPropertyOrNull("referenced_message")?.Pipe(Message.Parse);
 
             var content = type switch
             {
@@ -138,7 +144,8 @@ namespace DiscordChatExporter.Domain.Discord.Models
                 embeds,
                 reactions,
                 mentionedUsers,
-                messageReference
+                messageReference,
+                referencedMessage
             );
         }
     }
--- a/DiscordChatExporter.Domain/Discord/Models/Message.cs
+++ b/DiscordChatExporter.Domain/Discord/Models/Message.cs
@@ -48,6 +48,8 @@ namespace DiscordChatExporter.Domain.Discord.Models
 
         public IReadOnlyList<User> MentionedUsers { get; }
 
+        public MessageReference? Reference {get; }
+
         public Message(
             string id,
             MessageType type,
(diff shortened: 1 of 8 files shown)


Sanqui:
Previously:
image

Now:
image

Tyrrrz:
Thanks for the PR :)
I will take a look on ~Monday.

Tyrrrz:
Briefly looked through the HTML code just now. Can you please make it so that the reply quote block also links to the referenced message? There is a function called `scrollToMessage(event, messageId)` that you can use. Here's an example:

https://github.com/Tyrrrz/DiscordChatExporter/blob/d8dc12b3f758c9180579e41309c3e63f041d6d92/DiscordChatExporter.Domain/Exporting/Writers/MarkdownVisitors/HtmlMarkdownVisitor.cs#L143-L146

DiscordChatExporter.Domain/Discord/Models/Message.cs
```diff
@@ -90,6 +99,8 @@ public static Message Parse(JsonElement json)
             var callEndedTimestamp = json.GetPropertyOrNull("call")?.GetPropertyOrNull("ended_timestamp")?.GetDateTimeOffset();
             var type = (MessageType) json.GetProperty("type").GetInt32();
             var isPinned = json.GetPropertyOrNull("pinned")?.GetBoolean() ?? false;
+            var messageReference = json.GetPropertyOrNull("message_reference")?.Pipe(MessageReference.Parse);
+            var referencedMessage = json.GetPropertyOrNull("referenced_message")?.Pipe(Message.Parse);
```
Tyrrrz, line 103: Do we need both `message_reference` and `referenced_message`?
From the docs, it seems that `referenced_message` will always be present in replies, unless the backend didn't fetch it, in which case we may as well pretend that there is no referenced message.

image

DiscordChatExporter.Domain/Discord/Models/Message.cs
```diff
@@ -48,6 +49,10 @@ public partial class Message : IHasId
 
         public IReadOnlyList<User> MentionedUsers { get; }
 
+        public MessageReference? Reference {get; }
+
+        public Message? ReferencedMessage {get; }
```
Tyrrrz, line 54: Nit: space after `{` -> `{ get; }`

DiscordChatExporter.Domain/Discord/Models/MessageReference.cs
```diff
@@ -0,0 +1,37 @@
...
+    // https://discord.com/developers/docs/resources/channel#message-object-message-reference-structure
+    public partial class MessageReference
+    {
+        public string? MessageId { get; }
+
+        public string? ChannelId { get; }
+
+        public string? GuildId { get; }
+
+        public MessageReference(string? message_id, string? channel_id, string? guild_id)
```
Tyrrrz, line 16: Naming convention: `messageId` instead of `message_id` (and other fields too)

DiscordChatExporter.Domain/Discord/Models/MessageReference.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    // reference data sent with crossposted messages and replies
+    // https://discord.com/developers/docs/resources/channel#message-object-message-reference-structure
+    public partial class MessageReference
+    {
+        public string? MessageId { get; }
+
+        public string? ChannelId { get; }
+
+        public string? GuildId { get; }
```
Tyrrrz, line 14: Discord API docs claim that all of these can be nullable, but it's a bit weird to me. I think we should be able to assume that these are actually not nullable.

DiscordChatExporter.Domain/Discord/Models/MessageReference.cs
```diff
@@ -0,0 +1,37 @@
+using System.Text.Json;
+using JsonExtensions.Reading;
+
+namespace DiscordChatExporter.Domain.Discord.Models
+{
+    // reference data sent with crossposted messages and replies
+    // https://discord.com/developers/docs/resources/channel#message-object-message-reference-structure
+    public partial class MessageReference
```
Tyrrrz, line 8: Note the comment on Message:102, we might not need this class at all.

DiscordChatExporter.Domain/Exporting/Writers/JsonMessageWriter.cs
```diff
@@ -261,6 +261,17 @@ public override async ValueTask WriteMessageAsync(Message message)
...
+            // Reference
+
+            if (message.Reference is not null)
+            {
+                _writer.WriteStartObject("reference");
+                _writer.WriteString("messageId", message.Reference.MessageId);
+                _writer.WriteString("channelId", message.Reference.ChannelId);
+                _writer.WriteString("guildId", message.Reference.GuildId);
+                _writer.WriteEndObject();
+            }
```
Tyrrrz, line 273: We could probably render the whole message object here, given that we have it.

DiscordChatExporter.Domain/Exporting/Writers/Html/MessageGroup.cs
```diff
@@ -14,10 +14,21 @@ internal partial class MessageGroup
 
         public IReadOnlyList<Message> Messages { get; }
 
-        public MessageGroup(User author, DateTimeOffset timestamp, IReadOnlyList<Message> messages)
+        public MessageReference? Reference { get; }
```
Tyrrrz, line 17: Looks like only `ReferencedMessage` is used, but not `Reference`.

Tyrrrz (changes requested):
Looks good! Left a few comments. Maybe we can simplify it a bit.

DiscordChatExporter.Domain/Discord/Models/Message.cs
Sanqui: >in which case we may as well pretend that there is no referenced message.

Disagreed, if the message wasn't fetched (which could also happen if it was deleted), you still want to preserve the information that this message was a reply for context.  Otherwise you could get a discussion that reads in a misleading way.  Discord itself seem to convey it like this:
image

Tyrrrz: Ah, I see. I didn't account for deleted messages. Makes sense.

Tyrrrz:
Thanks for the PR! I will merge this to expedite the process as I want to release this week.

Sanqui:
Apologies that I haven't gotten back to this, I've ended up a little busy.

Tyrrrz:
No worries

PR #459, opened 2020-12-17
Set last write time on downloaded media to the value provided in the …

### LINKED ISSUE #418: Preserving upload dates on downloaded media
this app has been fantastic for archiving discord content, and i've been using it for a long while now. the new ability to download all the attachments has been super great, but a feature it lacks is preserving the "file modified" date on the downloaded media files. i'd find this super useful as i'm in a few discord servers where there are channels specifically for uploading game saves as attachments, and it would help out a lot if i could download all of these at once using this app, and then be able to have a whole archive of all of these game saves with all of their dates intact.

the attachment files hosted on discord do appear to have accurate file modified dates intact, but you can play it safe (just incase discord stops preserving these dates) by dating the files based on the date/time of the message it was uploaded on.

andrewkolos:
Closes #418.

This change makes DCE preserve the last modified date on downloaded media. For example, if a file was created on December 1, posted on Discord at Dec 2, and downloaded by DCE on Dec 3, the last-modified date of the created/downloaded file on the local machine will be Dec 1 (exact times omitted for brevity).

My lazy self didn't measure the potential performance impact of setting the last modified time. I think this adds a second file opening/closing, but I couldn't find a way keep this at one open/close without getting messy and calling into winAPIs.

### DIFF
--- a/DiscordChatExporter.Domain/Internal/Extensions/HttpClientExtensions.cs
+++ b/DiscordChatExporter.Domain/Internal/Extensions/HttpClientExtensions.cs
@@ -1,4 +1,7 @@
-﻿using System.IO;
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Thanks!

PR #462, opened 2020-12-19
[CLI] Allow passing after and before message ids instead of dates

### LINKED ISSUE #456: Allow passing after and before message ids instead of dates
It would be great if we could say "take all messages after the one with this id and until the one with this id", it would allow for more precision and it would also be way more practical

andrewkolos:
Closes #456.


Tyrrrz:
> Allow for retrieval of message by their ID, to get it's post time

You don't actually need to, the ID is already a timestamp. The `after` and `before` in Discord API take message IDs.

andrewkolos:
> > Allow for retrieval of message by their ID, to get it's post time
> 
> You don't actually need to, the ID is already a timestamp. The `after` and `before` in Discord API take message IDs.

Oh yeah, that's a great point.

andrewkolos:
What would we like to have happen in the case where there is no message that has the message ID found in `--after`/`--before`? We could just continue as normally, not caring whether or not the message actually exists.

Instead, we could throw an error, but this would require more work. We would need to detect that the provided ID doesn't have a corresponding message. Off the top of my head, one of the following would work:

* The `After` and `Before` members of `ExportRequest` need to include additional information about whether the datetimes were given directly from the user or if they were derived from a message ID. If the latter, downstream code checks to see the messages were encountered.
* When given IDs for `--after`/`--before`, the CLI calls the Discord API to see if these messages actually exist. This seems the simplest but would require a network call.
* Downstream code provides a means to listen to its progress. Specifically, it includes information about messages it encounters. This could take the form of `DiscordClient.GetMessagesAsync` including message IDs in its progress reporting through the `progress` argument. A big problem is that we would have to wait for the entire process to finish before we get to see if the last message is actually there.

I personally prefer the second option. Let me know what you think.

Tyrrrz:
@andrewkolos The closest would be okay. If I remember correctly, that's what happens anyway in Discord API. So take the path of least work :)

andrewkolos:
Should be ready to go. Interestingly, the --after message is included, but the --before message is not. This appears to be consistent with how the two options work when dates are given. Is this intended?

Tyrrrz:
@andrewkolos I think this is how Discord handles it.

Tyrrrz:
Thanks!

Tyrrrz:
@andrewkolos I added a `Snowflake` type so now these kind of manipulations should be easy. Now you can just do this:

https://github.com/Tyrrrz/DiscordChatExporter/blob/3d9ee3b3391ef8fccac8ded08dae626d3ca5d2c1/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs#L21-L25

PR #472, opened 2021-01-06
Formatting output paths

96-LB:
Currently, if you export multiple channels, there is no way to specify the filenames of each individual channel. Consider the following guild:
image

When these channels are exported with using either the `exportguild` CLI command or multi-selection on the GUI, the output directory appears as follows:
image
This is often not desirable - after all, that's why the option is given to rename exports when done one-by-one. But if exporting multiple channels, the user must rename them manually or export them individually.

This pull request seeks to change that by adding a new feature, treating output paths as formattable strings. Whether exporting a single channel or several, using certain escape sequences in the file/directory names enables the user to query metadata and insert them into the filepath. This can be used to name files in any desired format:
image

It can also be used to place the files in different directories and control the directory structure:
image

These formatting options make it much more convenient to back up a large number of channels and maintain a desired organization, as well as supporting easier automation. To use format flags, simply include one of the following strings in the output path:
```asciidoc
%g => Guild ID
%G => Guild Name
%t => Category ID
%T => Category Name
%c => Channel ID
%C => Channel Name
%p => Channel Position (0-indexed)
%P => Category Position (0-indexed)
%a => "After" Date (defaults to start of Discord era)
%b => "Before" Date (defaults to current day)
%% => "%" Literal
```
(These are probably some pretty poor choices for format strings, but they are easily configurable. Additional suggestions include 1-indexed positions and channel position within category; that last one may be a lot more work but I really want it😟)

For example, the string `%G - %g/#%C - %%%c%.html` maps to `96 LB's server_test_ - 796191206379225109/#general - %796191206379225111%.html`. As usual, if the specified path is a directory, the files will be given their default file name and will be placed in that directory. This could produce unexpected results if your string somehow ends in a file extension post-formatting.

Some additional notes about this pull request:
* In order to access category data, `Channel.Category` was refactored from `string?` to `Channel?`
* In order to render the new categories in the GUI, `Channel` implements IComparable and is sorted by order in Discord
* For both of the above reasons, `int Channel.Position` now exists
* As a side effect, sorting in the GUI now takes place by position. I think this is an improvement over alphabetical, and there exist a couple open issues pertaining to the sorting, but as they request a toggle I'm not going to close them.
* This feature is mainly useful in the CLI, however it is accessible in the GUI as well: one simply needs to select folder names that include formatting flags. These folders will remain unmodified, acting as "template" folders, while new ones are created. If a folder name ends in a file extension such as `.html`, the formatted output will be treated as a file instead of a directory. The formatting behaviour is somewhat unexpected, and I can't change it because there's no filepath option on the GUI (the user is instead prompted with a folder-only file dialog), so this PR may not yet be ready for merging if this is undesired.
* It's past midnight so I probably forgot something lol


96-LB:
Some more notes:
* Looks like the channels within each category are still sorted alphabetically since some other part of the code is overriding the default sort
* "Default" categories (e.g. "Text") are assigned a position value of -1
* It may be worth allowing leading zeroes in position values

cancerousname69:
I believe that this would be a great addition.

DiscordChatExporter.Domain/Discord/DiscordClient.cs
```diff
@@ -119,9 +119,10 @@ public async IAsyncEnumerable<Channel> GetGuildChannelsAsync(Snowflake guildId)
 
                 var categories = response
                     .EnumerateArray()
+                    .Where(j => j.GetProperty("type").GetInt32() == (int)ChannelType.GuildCategory)
                     .ToDictionary(
                         j => j.GetProperty("id").GetString(),
-                        j => j.GetProperty("name").GetString()
+                        j => Channel.Parse(j)
                     );
```
Tyrrrz, line 126: Can probably do `.EnumerateArray().Select(Channel.Parse)` to avoid parsing JSON directly. Since we do it later anyway.

DiscordChatExporter.Domain/Discord/DiscordClient.cs
```diff
@@ -162,10 +163,10 @@ public async IAsyncEnumerable<Role> GetGuildRolesAsync(Snowflake guildId)
             return response?.Pipe(Member.Parse);
         }
 
-        private async ValueTask<string> GetChannelCategoryAsync(Snowflake channelParentId)
+        private async ValueTask<Channel> GetChannelCategoryAsync(Snowflake channelParentId)
         {
             var response = await GetJsonResponseAsync($"channels/{channelParentId}");
-            return response.GetProperty("name").GetString();
+            return Channel.Parse(response);
         }
```
Tyrrrz, line 170: Isn't this the same as `GetChannelAsync` at this point?

DiscordChatExporter.Domain/Exporting/ExportRequest.cs
```diff
@@ -81,6 +83,27 @@ public partial class ExportRequest
             Snowflake? after = null,
             Snowflake? before = null)
         {
+
+            // Formats path
+            outputPath = Regex.Replace(outputPath, "%.", delegate (Match m)
```
Tyrrrz, line 88:
```suggestion
            outputPath = Regex.Replace(outputPath, "%.", m =>
```

DiscordChatExporter.Domain/Discord/Models/Channel.cs
```diff
@@ -36,44 +37,67 @@ public partial class Channel : IHasId
 
         public Snowflake GuildId { get; }
 
-        public string Category { get; }
+        public Channel? Category { get; }
```
Tyrrrz, line 40: This should never be nullable because we have `GetDefaultCategory`

DiscordChatExporter.Domain/Discord/DiscordClient.cs
96-LB: If I'm understanding correctly, this fully parses every channel regardless of whether it's a category or not. Would it be worth enumerating `response` only once and grouping everything by type through the JSON? Then we could later parse categories once and then afterwards text channels once.

DiscordChatExporter.Domain/Discord/Models/Channel.cs
96-LB: This brings up an issue: since categories are also channels, what should the category of a default category be? Previously I had used null, as using `GetDefaultCategory` would create infinite recursion. One thought is to modify the `Channel` constructor to set a category's category to itself if none is provided, but that feels hacky.

96-LB:
I made a few minor changes to the code based on your feedback. I'm not too convinced about my "fix" to allow changing `Channel.Category` to be non-nullable, so if you have a better solution, let me know.

A separate thought: I took a look through the Discord API and it seems that channel positions aren't actually guaranteed to be contiguous or even unique (sorting in the client happens first by position and then by id). This may produce some odd results if one wants to use the channel position flags, and I'm wondering if it's possible to determine channel positions based on their actual order instead of Discord's raw positions.

DiscordChatExporter.Domain/Discord/DiscordClient.cs
Tyrrrz: Parsing itself is not an expensive operation (compared to network i/o) so if it helps us avoid duplication (i.e. referring to field names in multiple places) then it's worth it.

DiscordChatExporter.Domain/Discord/Models/Channel.cs
```diff
@@ -36,44 +37,67 @@ public partial class Channel : IHasId
...
         public string? Topic { get; }
 
-        public Channel(Snowflake id, ChannelType type, Snowflake guildId, string category, string name, string? topic)
+        public Channel(Snowflake id, ChannelType type, Snowflake guildId, Channel? category, string name, int position, string? topic)
         {
             Id = id;
             Type = type;
             GuildId = guildId;
-            Category = category;
+            Category = category ?? (type != ChannelType.GuildCategory ? GetDefaultCategory(type, guildId) : this);
```
Tyrrrz, line 53: That's a good point, I didn't think of this.

I can see a couple of solutions to this:

1. Keep `Category` nullable, but then we have to be a bit more defensive in the output writers, as we can't rely on the category being set.
2. Create a `ChannelCategory` class, which is a `Channel` with only properties relevant to a category. So it wouldn't have a parent category of its own (IIRC that's not possible, right?) and topic.

And finally, the greatest of them all:

3. Do we really need the expanded category metadata? It seems what we're using it for is: category position and category ID. As a Discord user, I think, you're not even aware that the categories are actual channels rather than simple labels, so perhaps those are not very useful. I personally have a hard time imagining where I would want the category ID or position to be included. What do you think? If we can drop it, we can revert the changes to category parsing and make it significantly simpler.
3.

Tyrrrz (commented):
Thanks for the PR and review changes! Did another pass, this time putting a bit more thought into this feature.

Left a few new comments.

One additional thing I didn't notice initially, please add `.vs` directory to `.gitignore` and remove the `.vs/DiscordChatExporter/DesignTimeBuild/.dtbcache.v2` file. I personally don't use Visual Studio so didn't add it to gitignore, but it looks like it drags some temp files that we'd rather not commit.

DiscordChatExporter.Domain/Discord/Models/Channel.cs
Tyrrrz: Good point, I didn't realize that. Raised it in a separate discussion.

DiscordChatExporter.Domain/Exporting/ExportRequest.cs
```diff
@@ -81,6 +83,27 @@ public partial class ExportRequest
             Snowflake? after = null,
             Snowflake? before = null)
         {
+
+            // Formats path
+            outputPath = Regex.Replace(outputPath, "%.", m =>
```
Tyrrrz, line 88:
```suggestion
            outputPath = Regex.Replace(outputPath, "%\w", m =>
```

A bit less aggressive pattern. Putting catch-alls (`.`) scares me for some reason. :D

DiscordChatExporter.Domain/Discord/Models/Channel.cs
Tyrrrz: Approach `3.` may break the GUI sorting though, because then we wouldn't be able to sort the categories by position.

Just to clarify, `Channel.Position` is the channels global position within the guild, or is it within its category?

96-LB: I think you raise a valid point in that I was definitely attaching more data than needed. I think condensing the data into a few properties on the `Channel` as in approach `3.` is fine, unless it becomes too cluttered in which case it may be better to follow solution `2.` instead.

I think position is definitely a useful piece of data, since it lets you organize categories and channels in your filesystem the same way which they would appear in Discord. For example, I currently have a script that exports my servers like so:
image
ID is perhaps not as useful, but I could see users potentially wanting to access that.

As for the actual value of `Channel.Position`...well, I originally intended it to be its position within the category, but Discord seemed to return global position through the API so I just went with that. I've done some more digging since making the initial PR proposal and it seems that the `position` value which Discord attaches to each channel isn't absolute, without any guarantees that the positions will be unique or contiguous. It's not very clear on the documentation, but there's some discussion about it [here](https://github.com/Rapptz/discord.py/issues/2392). As such, using Discord's raw response may not actually be useful at all. I think it's definitely possible to assign position values (both within category and global) when enumerating the channels, but I believe that would result in an extra request for the CLI if exporting a single channel (though, in that case, maybe channel position isn't very important). However, things might get a bit messy if we stuff in manual data that can't be parsed with `Channel.Parse`. Thoughts?

DiscordChatExporter.Domain/Exporting/ExportRequest.cs
96-LB: This no longer catches `%%`. We could add a special case for that, but I don't think the catch-all is a problem since most patterns are simply ignored due to the `_ => m.Value` case.

96-LB:
Whoops, sorry about that! Visual Studio kept trying to push unhelpful changes and files and that one must have slipped by. I've responded to your comments, and I'll probably be able to make some edits tomorrow.

DiscordChatExporter.Domain/Discord/Models/Channel.cs
```diff
@@ -21,7 +22,7 @@ public enum ChannelType
     }
 
     // https://discord.com/developers/docs/resources/channel#channel-object
-    public partial class Channel : IHasId
+    public partial class Channel : IHasId, IComparable, IComparable<Channel>
```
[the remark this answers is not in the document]
96-LB: I pulled it out, but I couldn't figure out how to apply it to the XAML, which kept yelling at me that Channel didn't implement IComparable. The extent of my knowledge regarding WPF is everything I googled in the past 30 minutes, so in the end I just modified one of the sorts to use the Position property instead of Category, and that appears to be correctly sorting everything according to the Discord hierarchy. I don't know if the comparer class is still needed (or even if what I did is correct), so I'll await your input on that.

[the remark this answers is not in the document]
Tyrrrz: Yeah, don't worry about that, I will fix it if it's broken later. It's not the main goal of the PR.

DiscordChatExporter.Domain/Discord/Models/Channel.cs
Tyrrrz: I see, category position does appear useful.
Maybe we can create a special type, `ChannelCategory` which would just contain the name, the position, and the ID? I don't know off the top of my head if that helps simplify things, but at least there will be no referential recursion.

DiscordChatExporter.Domain/Exporting/ExportRequest.cs
Tyrrrz: Ah, I see. Good point.

DiscordChatExporter.Domain/Discord/Models/Channel.cs
96-LB: Sounds good. Since the type would overlap with `Channel`, do you think it would be better to use some sort of inheritance or keep them completely separate?

Tyrrrz: Let's keep them completely separate to leave our options open for the future.

96-LB:
In that last commit I took a stab at correcting the position values given by Discord. Essentially, I sorted and enumerated the channels and categories before parsing them, and replaced the given position value with the index after sorting. This fixes any non-unique or non-contiguous positions values, making `Channel.Position` refer to the _actual_ order (guild-wide, with separate counts for categories and channels) instead of Discord's raw position value. Additionally, I changed default categories to be category 0 instead of -1.

Here's an example of old vs. new behaviour on a newly created server:
image

The logic is currently handled in `DiscordClient.GetGuildChannelsAsync`, so I think it misses the CLI's export-single-channel command since the `Channel` object is created before the context is grabbed. This commit can be easily extended to also capture a channel's position within its category (which I believe is probably even more useful than global position), but I decided to hold off on that until you took a look.

Tyrrrz:
Thanks! 🎉

96-LB:
Small bug introduced by this PR: the CLI's `dm` command throws a runtime error since `Channel.Category` is no longer sortable.

Tyrrrz:
@96-LB I see. Do you have time to submit a PR to fix that?

96-LB:
Done!

PR #490, opened 2021-01-30
[CLI] Fix sorting by channel/category position

96-LB:
Fixes a bug introduced by #472. The unused `ChannelBasedPositionComparer` was replaced with a more generic `PositionBasedComparer`, which now correctly sorts first by position and then by ID. `IHasIdAndPosition : IHasId` was created to facilitate this, and applied to both `Channel` and `ChannelCategory`. The new comparer was used in the CLI's `dm` and `channels` commands to stop them from throwing exceptions.


Tyrrrz:
Thanks!

PR #497, opened 2021-02-07
Add partition by file size

### LINKED ISSUE #477: Add partition by file size
### Tl;dr: add ability to partition output files by file size.

_You can really tell I wrote this at 4 in the morning, haha:_

Being able to partition large files at a specified file size would be a useful addition. Partitioning by message number is good, but depending on the content of the channel, file sizes of the output can vary significantly, and getting the "right" size can take a bit of guesswork. Messages in some channels—such as those used by a bot to post server events like user joins—are basically all the same size, but user off-topic channel message sizes are all over the place, with the result that one channel can produce files which are double the size of another channel. Setting a file size rather than a message count would take all the guesswork out of it, and produce similarly sized files in one batch without having to run the tool with different settings for different channels.

andrewkolos:
Closes #477.

- [x] Need to adjust GUI code to support the feature.
- [x] Add handling of bad input for `--partition`.
- [x] Still needs testing.

`--partition` now accepts file sizes. Examples: 1MB, 2 mb, 0.5 mB, 0.3 Mb. When the current file stream exceeds the partition limit, the next partition is created.

This implementation doesn't absolutely cap the file size to the maximum -- it won't create a new partition until we have actually exceeded the maximum size (I should have added clarification in the CommandOption description, but I couldn't think of a concise way to indicate this).

~~For the GUI, this could be supported by adding a combination of a dropdown/select for partition type and then a numeric text entry for the value, or we could just limit the GUI to partitioning by message. I'll try the former when I get the time.~~ edit: added support to GUI; provided value is always interpreted as an amount of megabytes

image


Tyrrrz:
Thanks!

Now that I think about it, it may be possible that the user would want to partition by both file size and messages (i.e. smaller than 10mb and no more than 1000 messages). I don't know how likely that is, considering that the main goal of partitioning is to limit the file size, which the file size option obviously does well, to the point that the other option may not be necessary.

Basically, I think it's either of the two possible options:

- The user will only use partitioning to make sure the files are not too big for the browser/editor to open. This is handled by the file size option in full capacity, effectively making the other option obsolete. If that's the case, we may want to make it obsolete and remove in the future.
- The user may want to use partitioning for something else. In this case, it's likely they may want to use both file size partitioning and message partitioning at the same time. This means, both options must be available side-by-side.

@dziban303
As the author of the original issue, what are your thoughts on this?

dziban303:
The first bullet was my reasoning for requesting the addition. As a server admin, I find that, with some regularity, I have to refer back to something that's happened a few years back. Discord search is just awful so having the logs available locally is **great**. I grep around in the directory until I find what I'm looking for, and open the html file. Unfortunately, the browser creaks loudly when trying to load a 80MB+ html file; likewise, neither Sublime or Notepad++ are pleased. Partitioning by message count would result in sensible file sizes for the active channels, but lots of tiny files for the less active channels.

As for the second bullet point, I don't need for that for my use case, but I certainly can't say someone else won't find that to be a huge help.

Tyrrrz:
> Partitioning by message count would result in sensible file sizes for the active channels, but lots of tiny files for the less active channels.

Can you explain why that is? Partitioning by message count should work the same way across both active and less active channels, since it just splits by certain number of messages, not by how often they're posted? The only case I can think of that could matter is if one channel is more likely to have much larger messages than another channel, but that's not necessarily dependent on how active the channel is.

dziban303:
Compare a bot channel which consist of, basically, “abc#123 has joined the server,” versus (in our particular case) weather analysis channels with frequent messages of 500-1000 characters and heavy with graphics and links. Another server I like to back up deals with discussing books, with quite a bit of messages being larger than what I presume is a Discord-wide average value. From: Alexey GolubSent: Wednesday, March 24, 2021 16:09To: Tyrrrz/DiscordChatExporterCc: Dziban303; MentionSubject: Re: [Tyrrrz/DiscordChatExporter] Add partition by file size (#497) Partitioning by message count would result in sensible file sizes for the active channels, but lots of tiny files for the less active channels.Can you explain why that is? Partitioning by message count should work the same way across both active and less active channels, since it just splits by certain number of messages, not by how often they're posted? The only case I can think of that could matter is if one channel is more likely to have much larger messages than another channel, but that's not necessarily dependent on how active the channel is.—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

Tyrrrz:
@dziban303 
Understood. To summarize, you feel like partitioning by file size is enough and also makes partitioning by message count unnecessary (for your use cases).

andrewkolos:
Hey all. Sorry--I've been busy hunting for a job. I've read through this discussion and it seems like this addition is desired and usable as is, right? I'm happy to catch this back up with the main branch.

dziban303:
For my part I'm quite happy with it and I appreciate the work put in to
accommodate my request. Thanks!

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -24,8 +24,9 @@ public abstract class ExportCommandBase : TokenCommandBase
         [CommandOption("before", Description = "Only include messages sent before this date or message ID.")]
         public Snowflake? Before { get; init; }
 
-        [CommandOption("partition", 'p', Description = "Split output into partitions limited to this number of messages.")]
-        public int? PartitionLimit { get; init; }
+        [CommandOption("partition", 'p', Description = "Split output into partitions limited to this number of messages or a maximum file size (e.g. \"25mb\").",
+            Validators = new[] { typeof(PartitionValidator) })]
+        public string? PartitionLimit { get; init; }
```
Tyrrrz, line 29: What I was thinking was to use `IPartitioner` right here, instead of `string`. We can do it either by providing a converter, or by changing the interface to an abstract class and adding `Partitioner.Parse(string)` method.

DiscordChatExporter.Domain/Exporting/MessageExporter.cs
```diff
@@ -1,29 +1,37 @@
...
-            _messageCount > 0 &&
-            _context.Request.PartitionLimit != null &&
-            _context.Request.PartitionLimit != 0 &&
-            _messageCount % _context.Request.PartitionLimit == 0;
+        private bool IsPartitionLimitReached()
+        {
+            if (_writer == null)
+            {
+                throw new Exception("Cannot check for partition limit being reached when writer is null.");
+            }
```
Tyrrrz, line 32: Can just return `false` in this case I think

DiscordChatExporter.Domain/Exporting/Partitioners/IPartitioner.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DiscordChatExporter.Domain.Exporting
+{
+    interface IPartitioner
+    {
+        bool IsLimitReached(long messageCount, long sizeInBytes);
```
Tyrrrz, line 9: Can we make it `bool IsLimitReached(ExportPartitioningContext context)` and move the properties to a class?

DiscordChatExporter.Gui/Internal/PartitionFormat.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DiscordChatExporter.Gui.Internal
+{
+    public enum PartitionFormat
+    {
+        MessageCount,
+        Filesize,
```
Tyrrrz, line 10:
```suggestion
        FileSize,
```

Tyrrrz:
@andrewkolos left a few comments.
Let's get this in and I'll refactor a bit :)

andrewkolos:
@Tyrrrz Was `ArgumentValueValidator` removed in a newer release of CliFx? I can't find it.

Tyrrrz:
@andrewkolos it's now called `BindingValidator`.

DiscordChatExporter.Domain/DiscordChatExporter.Domain.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="ByteSize" Version="2.0.0" />
```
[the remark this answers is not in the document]
andrewkolos: Sure, how'd you like it to work? Would it use the `TryParse(s, out result)` pattern, an enum pattern with success and failure objects (e.g. `FileSizeParseResult.Ok` and `FileSizeParseResult.Error`), or would you like a single FileSizeParseError class, where a null value indicates success, like how CliFx does it now? 

it might be simpler for you to do this yourself real quick once things are ready.

DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
```diff
@@ -236,6 +237,25 @@ public async void ExportChannels()
...
+            IPartitioner CreatePartitioner()
+            {
+                var partitionFormat = dialog.SelectedPartitionFormat;
+                var partitionLimit = dialog.PartitionLimit;
+                if (partitionLimit == null)
+                {
+                    return new NullPartitioner();
+                }
+
+                if (partitionFormat == PartitionFormat.MessageCount)
```
andrewkolos, line 250: Maybe `PartitionFormat` should implement an enum class pattern and have a `CreatePartitioner` method to avoid this if-else chain. However, if we don't plan on adding any more partition formats, this seems like extra work for nothing.

DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
```diff
@@ -236,6 +237,25 @@ public async void ExportChannels()
             // Notify of overall completion
             if (successfulExportCount > 0)
                 Notifications.Enqueue($"Successfully exported {successfulExportCount} channel(s)");
+
+            IPartitioner CreatePartitioner()
```
andrewkolos, line 241: Maybe `PartitionFormat` should implement an enum class pattern and have a `CreatePartitioner` method to avoid this if-else chain. However, if we don't plan on adding any more partition formats, this seems like extra work for nothing.

Also, why do I need these `(int)` casts near the end? They are guarded by a previous if-return that checks if `partitionLimit` is null. I thought the C# compiler could recognize this?

andrewkolos: Sorry, I thought I selected every line of `CreatePartitoner` to mention in this comment, but I guess I don't know how to do that XD

andrewkolos:
Not sure why the build is failing, I don't have Offline Packages selected as a source for any of the projects.

Tyrrrz:
> Not sure why the build is failing, I don't have Offline Packages selected as a source for any of the projects.

It's a GitHub Actions infrastructure issue 🙄 
Don't worry about it.

DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
Tyrrrz: > Also, why do I need these (int) casts near the end? They are guarded by a previous if-return that checks if partitionLimit is null. I thought the C# compiler could recognize this?

C# compiler is generally pretty dumb. It doesn't have the level of flow analysis that TS has, for example.

> Sorry, I thought I selected every line of CreatePartitoner to mention in this comment, but I guess I don't know how to do that XD

NP. To do it, you need to click and hold.

DiscordChatExporter.Cli/Commands/Base/PartitionValidator.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class PartitionValidator : BindingValidator<string>
+    {
+        public override BindingValidationError? Validate(string value)
+        {
+            return int.TryParse(value, out _) ? Ok() :
+                ByteSize.TryParse(value, out _) ?
+                Ok() :
+                Error(@"The value must be numeric or represent a file size (e.g. 25mb).");
+        }
+    }
```
Tyrrrz, line 20: I now realize it's not immediately obvious, but `BindingValidator` is triggered after the value is already converted, so this should be `BindingValidator<IPartitioner>` instead. However, seeing as the validation is in a way already handled in `BindingConverter`, I think this validator is not necessary at all.

Tyrrrz: Created https://github.com/Tyrrrz/CliFx/issues/103

DiscordChatExporter.Core/Exporting/MessageExporter.cs
```diff
@@ -38,7 +48,7 @@ private async ValueTask ResetWriterAsync()
         private async ValueTask<MessageWriter> GetWriterAsync()
         {
             // Ensure partition limit has not been exceeded
-            if (IsPartitionLimitReached())
+            if (_writer != null && IsPartitionLimitReached())
```
Tyrrrz, line 51:
```suggestion
            if (_writer is not null && IsPartitionLimitReached())
```

For consistency

DiscordChatExporter.Core/Exporting/Partitioners/IPartitioner.cs
```diff
@@ -0,0 +1,12 @@
+﻿using DiscordChatExporter.Core.Exporting.Partitioners;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DiscordChatExporter.Core.Exporting
+{
+    public interface IPartitioner
+    {
+        public bool IsLimitReached(ExportPartitioningContext context);
```
Tyrrrz, line 10:
```suggestion
        bool IsLimitReached(ExportPartitioningContext context);
```

Adding access modifiers in interface members has a special meaning -- it denotes a default interface implementation. It has no effect here because the method doesn't have a body, but generally it's better to omit access modifiers in interfaces.

DiscordChatExporter.Gui/Views/Dialogs/ExportSetupView.xaml.cs
```diff
@@ -1,4 +1,8 @@
-﻿namespace DiscordChatExporter.Gui.Views.Dialogs
+﻿using DiscordChatExporter.Gui.Internal;
+using System;
+using System.Windows;
+
+namespace DiscordChatExporter.Gui.Views.Dialogs
```
Tyrrrz, line 5:
```suggestion
﻿namespace DiscordChatExporter.Gui.Views.Dialogs
```

Looks like these were added with some other changes, but are no longer needed

DiscordChatExporter.Gui/ViewModels/RootViewModel.cs
Tyrrrz: You can, however, streamline it like this:

```csharp
return (partitionFormat, partitionLimit) switch
{
   (PartitionFormat.MessageCount, int messageLimit) => new MessageCountPartitioner(messageLimit),
   (PartitionFormat.FileSize, int fileSizeLimit) => new FileSizePartitioner(fileSizeLimit),
   _ => new NullPartitioner()
}
```

(my syntax may be wrong, don't exactly remember how to write this from memory)

andrewkolos: Very cool

Tyrrrz:
Thanks! I'll take it from here.

PR #499, opened 2021-02-14
Skip JSON Validation

### LINKED ISSUE #413: Error occurred ']' invalid without a matching open
grafik

Why I think this happened:
grafik

( No, the harddisk is not only full with DC requests, it is only about 1GB.
But I had moved a few files at the same time and apparently there was not much space. )

But what I would say about the previous bug, it would be nice if there were some user friendly bugs, so for example it just pauses instead of crashing.

96-LB:
Closes #413 

Tiny modification to turn off validation of JSON output as discussed in the linked issue. The motivation is to prevent exceptions from being thrown during disposal of the JSON writer and hiding other exceptions.


Tyrrrz:
Thanks!

PR #524, opened 2021-03-18
Fix JSON discriminator and nickname fields

### LINKED ISSUE #500: Wrong leading zero user discriminator in JSON
When using JSON as an export format, in mentions of users, if user has leading zeros in discriminator (zeros before the first nonzero digit) these discriminators are being written without these zeros which results in incorrect discriminator.

96-LB:
Closes #500 

Tiny fix to address discriminators with leading zeroes not printing correctly in mentions as well as missing nickname fields in author objects. It was suggested in the linked issue to create a method to handle outputting user objects to the JSON since the code in the two locations is nearly identical (and that was the cause of the issue), but I decided to wait until further input was given before moving forward with that. The differences in the two areas are: 1. mention objects have no property name since they reside in an array and 2. `avatarUrl` is added to author objects but not mentions. Is this enough to warrant separation of the code, or could a method be made with some parameters to handle those differences?

### DIFF
--- a/DiscordChatExporter.Core/Discord/Data/User.cs
+++ b/DiscordChatExporter.Core/Discord/Data/User.cs
@@ -15,9 +15,11 @@ namespace DiscordChatExporter.Core.Discord.Data
 
         public int Discriminator { get; }
 
+        public string DiscriminatorFormatted => $"{Discriminator:0000}";
+
         public string Name { get; }
 
-        public string FullName => $"{Name}#{Discriminator:0000}";
+        public string FullName => $"{Name}#{DiscriminatorFormatted}";
 
         public string AvatarUrl { get; }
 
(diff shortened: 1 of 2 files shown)


Tyrrrz:
Thank you! 😃 

> It was suggested in the linked issue to create a method to handle outputting user objects to the JSON since the code in the two locations is nearly identical (and that was the cause of the issue), but I decided to wait until further input was given before moving forward with that.

I agree, I don't think it's necessary, at least now. If they end up being the same eventually, we can do it.

We could, however, add an evaluated property to `User` called `DiscriminatorFormatted` that converts it to a string. We are using that code in at least 3 different places and it could be nice to centralize it.

https://github.com/Tyrrrz/DiscordChatExporter/blob/7b140fd45081adbfc0cefa4f7a8d93390c3a4e5e/DiscordChatExporter.Core/Discord/Data/User.cs#L10-L22

96-LB:
Alright, I've patched that in.

Tyrrrz:
Thanks :)

PR #552, opened 2021-04-14
Write user color in JSON

rtm516:


Tyrrrz:
Just out of curiosity, what do you need this for?

rtm516:
Json is a much better archive format than HTML and I'm working on a web app that can load in the json files to render them out, also removed the markdown processing on the content field but figured that would likely not be wanted in here.

DiscordChatExporter.Core/Exporting/Writers/JsonMessageWriter.cs
```diff
@@ -230,6 +230,7 @@ public override async ValueTask WriteMessageAsync(Message message)
             _writer.WriteString("name", message.Author.Name);
             _writer.WriteString("discriminator", message.Author.DiscriminatorFormatted);
             _writer.WriteString("nickname", Context.TryGetMember(message.Author.Id)?.Nick ?? message.Author.Name);
+            _writer.WriteString("color", Context.TryGetUserColor(message.Author.Id)?.ToHex() ?? "#FFFFFF");
```
Tyrrrz, line 233: Let's not fall back to `"#FFFFFF"`  because that assumes we know what Discord would use a fallback color. And that's likely different depending on the theme.

Instead just write null if the color is null.
```suggestion
            _writer.WriteString("color", Context.TryGetUserColor(message.Author.Id)?.ToHex());
```

Tyrrrz:
Thanks!

PR #553, opened 2021-04-15
Add a raw json export format

rtm516:
This adds a JSON export format that doesn't parse any of the markdown just exports the raw data allowing for better post-processing outside the exporter.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportFormat.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportFormat.cs
@@ -8,7 +8,8 @@ namespace DiscordChatExporter.Core.Exporting
         HtmlDark,
         HtmlLight,
         Csv,
-        Json
+        Json,
+        JsonRaw
     }
 
     public static class ExportFormatExtensions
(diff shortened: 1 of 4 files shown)


DiscordChatExporter.Core/Exporting/ExportFormat.cs
```diff
@@ -30,6 +32,7 @@ public static class ExportFormatExtensions
             ExportFormat.HtmlLight => "HTML (Light)",
             ExportFormat.Csv => "CSV",
             ExportFormat.Json => "JSON",
+            ExportFormat.JsonRaw => "JSON (no-formatting)",
```
rtm516, line 35: This could probably have a better name.

Tyrrrz:
Thanks for the PR, but I don't want to add any more formats for now. Maybe in the future we'd add format-specific options and would allow some things configurable, but I doubt this is going to be one of them. JSON already doesn't do much to markdown anyway, it only processes mentions.

omegasome:
@Tyrrrz The fact that it processes those mentions is the problem, specifically when it deals with channel mentions. Threads consistently become `#deleted-channel` rather than resolving to the name of the thread, which means that that data is lost by the archive since it doesn't save the ID.

It also makes the JSON archive less useful _as_ a JSON archive, because it will record names with spaces which are impossible for other programs to parse: `Hey @Steve Smith did you see that new movie?` might be referencing `@Steve`, `@Steve Smith`, `@Steve Smith did`, or any number of other Steves.

PR #554, opened 2021-04-15
Export channel category id to JSON

rtm516:


Tyrrrz:
What's the use case?

rtm516:
It allows differentiation of channel categories if they have the same name.

PR #555, opened 2021-04-16
Fix emojis not being downloaded with json

rtm516:
This fixes an issue where emojis that are not used in reactions don't get downloaded with a JSON export.


Tyrrrz:
Media export only downloads referenced media, but emoji are not actually referenced in JSON export, so this is by design. Also you'd need to await `ResolveMediaUrlAsync(...)` inside an async method, which `VisitEmoji(...)` isn't.

PR #607, opened 2021-06-15
Render YouTube iframe for applicable video embeds

### LINKED ISSUE #570: Watch YouTube videos inside the HTML files
### Flavor

### Details

I have no idea if this is even possible, so this is partly a suggestion and partly a question if this is possible. As the title says, it's about watching YouTube videos directly in the HTML file just as you can in Discord itself:

image

quentinmay:
[Original Issue](https://github.com/Tyrrrz/DiscordChatExporter/issues/570)
Replace the typical youtube link "/watch?v=" with "/embed/" and instead of showing the video description, make it show the video iframe embed instead. 
This solution wouldn't work with shortened or different weird youtube URL formats, but works for the general case.

**Result**
image

Closes #570

### DIFF
--- a/DiscordChatExporter.Core/Exporting/Writers/Html/Core.css
+++ b/DiscordChatExporter.Core/Exporting/Writers/Html/Core.css
@@ -458,6 +458,10 @@ img {
     border-radius: 3px;
 }
 
+.chatlog__embed-youtube-container {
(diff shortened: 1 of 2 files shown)


Tyrrrz:
Thanks!

PR #622, opened 2021-06-26
Filter which messages are exported

### LINKED ISSUE #160: Export messages with filtering
Is there a way to export specific content such as only images or only videos from a channel?

96-LB:
Closes #160  
(See also #591)

This pull request introduces a new feature that allows the user to specify a filter that controls which messages are exported. Filters are created using a syntax similar to queries the Discord search bar, using `<filter>:<value>` syntax to denote a filter of type `<filter>` which consumes the value `<value>` (e.g., `from:1234567890` matches messages sent by the user with id "1234567890"). Standalone words without colons can also be used to filter messages which contain that text (e.g., `potato` matches messages which contain the string "potato"). In addition, some extra syntax is added to handle more complex queries:
* Filters can be negated by prepending a `-` (e.g., `-from:1234567890` instead excludes messages sent by that user)
* Filters can be stacked by concatenation (e.g., `from:1234567890 potato` matches messages that match both filters)
* Filters can be merged by infixing a `|` (e.g., `from:1234567890 | potato` matches messages that match either filter)
* Filters can be grouped with parentheses (e.g., `-(from:1234567890 | potato)` negates the entire filter specified above)
* Multiple words can be grouped into one token and special characters can be escaped with double quotes (e.g., `from:"96 LB#5274"` escapes the space and `":-)"` escapes the special characters)
* Double quotes and backslahes can be escaped with backslashes (e.g. `"this is\" one string"` and `"this is a backslash: \\"`)

The following filters are currently implemented:
* `contains:` - Matches any message which contains the specified string. This is what standalone text is implicitly converted into.
* `from:` - Matches any message sent by a user with a matching ID or username + discriminator combo.
* `has:` - Can be used to match messages with attachments of a certain type (`file`, `video`, `image`, `sound`, `embed`). ((TODO: implement `link`))
* `mentions:` - Matches any message which mentions a user with a matching ID or username + discriminator combo.

Whitespace between tokens is irrelevant except within quotes, and filter names are case-insensitive (e.g., `from :   123456790` is identical to `FroM:1234567890`). Filters can be set on the CLI through the `--filter` flag, and ((TODO)) on the GUI through the export popup.

<hr>

This PR is still a draft, but as it's a lot I wanted to make sure it was heading in the right direction. I followed pretty closely the styles used for DCE's partitioning as well as Superpower's sample programs, but if a different design pattern would work better, I'm willing to refactor. Here's some of the stuff that's unfinished or might need some edits:
- [x] Make accessible in GUI
- [x] Implement `has:link`
- [ ] Improve error message reporting
- [x] Tweak `contains:`/text token behaviour
- [ ] Probably a good idea to test some more

I'm also going to go through and put a few review comments on sections of the code. Feel free to point out anything additional that looks like an issue or could be improved.


DiscordChatExporter.Core/Exporting/ChannelExporter.cs
```diff
@@ -39,6 +39,10 @@ public async ValueTask ExportChannelAsync(ExportRequest request, IProgress<doubl
             var encounteredUsers = new HashSet<User>(IdBasedEqualityComparer.Instance);
             await foreach (var message in _discord.GetMessagesAsync(request.Channel.Id, request.After, request.Before, progress))
             {
+                // Skips any messages that fail to pass the supplied filter
+                if (!request.MessageFilter.Filter(message))
+                    continue;
+
```
96-LB, line 45: I chose to put the logic here instead of in the writer with the partitioning logic to avoid instantiating the writer/file if no messages are being exported (and the "exportedAnything" flag is here).

DiscordChatExporter.Core/Exporting/Filtering/ContainsMessageFilter.cs
```diff
@@ -0,0 +1,16 @@
...
+    public class ContainsMessageFilter : MessageFilter
+    {
+        private readonly string _value;
+
+        public ContainsMessageFilter(string value)
+        {
+            _value = value;
+        }
+
+        public override bool Filter(Message message) => message.Content.Contains(_value);
```
96-LB, line 14: This is a really primitive check that should probably be updated. Right now the token "a" will match any word with a lowercase letter a (e.g., it'll match the message "banana" but not the message "A"). We could try splitting the value and message content into words and making sure all words in the filter value are also in the message.

DiscordChatExporter.Core/Exporting/Filtering/FromMessageFilter.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class FromMessageFilter : MessageFilter
+    {
+        private readonly string _value;
+
+        public FromMessageFilter(string value)
+        {
+            ///TODO: should we validate input here?
+            _value = value;
+        }
```
96-LB, line 13: I decided against rejecting input that fails to look like a username+discriminator or a snowflake (right now it'll just fail to match anything), but I thought it might be better to throw an exception before starting the export.

DiscordChatExporter.Core/Exporting/Filtering/HasMessageFilter.cs
```diff
@@ -0,0 +1,28 @@
...
+        public HasMessageFilter(string value)
+        {
+            ///TODO: should we validate input here?
+            _value = value;
+        }
+
+        public override bool Filter(Message message) =>
+            _value switch
+            {
+                "link" => throw new System.NotImplementedException(), //how do we do this?
```
96-LB, line 19: I'm not actually 100% sure how Discord checks for links. I'd assume just some sort of regex on the message content?

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+        public abstract bool Filter(Message message);
+    }
+
+    public partial class MessageFilter
+    {
+        public static MessageFilter CreateFilter(string key, string value)
+        {
+            return key.ToLower() switch
+            {
+                "contains" => new ContainsMessageFilter(value),
```
96-LB, line 20: This could probably be dropped and replaced with an overload of MessageFilter.CreateFilter that only takes one parameter if we don't want the redundant `contrains:` filter type.

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+                "from" => new FromMessageFilter(value),
+                "has" => new HasMessageFilter(value),
+                "mentions" => new MentionsMessageFilter(value),
+                _ => throw new ArgumentException($"Invalid filter type '{key}'.", "key")
+            };
+        }
+
+        public static MessageFilter Parse(string value, IFormatProvider? formatProvider = null) =>
+            FilterParser.TryParse(value, out var filter, out var error, out var position) ? filter : throw new ParseException(error, position);
```
96-LB, line 29: I couldn't actually get these errors to show up in the console. They were being replaced with a generic message about being unable to set the `--filter` flag. Is CliFx masking them?

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterParser.cs
```diff
@@ -0,0 +1,76 @@
...
+                return false;
+            }
+
+            value = parsed.Value;
+            error = null;
+            errorPosition = Position.Empty;
+            return true;
+        }
+    }
+}
```
96-LB, line 76: This is my first time doing this parsing so I am extra open to suggestions on all the Superpower stuff. I haven't yet put in too much customization for the error messages, but see my other comment about the errors not being visible to the user anyway.

DiscordChatExporter.Core/Exporting/Filtering/HasMessageFilter.cs
Tyrrrz: I think we identify links currently when parsing for markdown, the links are parsed into instances of `LinkNode`.

DiscordChatExporter.Core/Exporting/Filtering/FromMessageFilter.cs
Tyrrrz: I think this is fine for now

DiscordChatExporter.Core/Exporting/Filtering/ContainsMessageFilter.cs
Tyrrrz: Just an idea, but maybe we should make it so the contains filter tries to match a word? So `a` would only match standalone `a` word (the article) separate by whitespace, but not as part of another word. Is that how Discord search works too, I wonder? In any case, this can be achieved by using `Regex.IsMatch(message.Content, "\\b" + Regex.Esacpe(_value) + "\\b")`

DiscordChatExporter.Core/Exporting/ChannelExporter.cs
Tyrrrz: Makes sense :+1:

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
Tyrrrz: Yeah, I think we should probably drop the redundant `contains:` to avoid confusion

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
Tyrrrz: I believe it should still include the original error message:

https://github.com/Tyrrrz/CliFx/blob/df73a0bfe80e4984e13a5f680cc5cd0382eead2e/CliFx/CommandBinder.cs#L164-L171

What is the error message that gets printed?

DiscordChatExporter.Core/Exporting/Filtering/MentionsMessageFilter.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        public MentionsMessageFilter(string value)
+        {
+            ///TODO: should we validate input here?
+            _value = value;
+        }
+
+        //match either a username + discriminator or an id
+        public override bool Filter(Message message) =>
+            message.MentionedUsers.Any(user => _value == user.FullName || _value == user.Id.ToString());
```
Tyrrrz, line 18: Maybe we should also try to match just the `user.Name`? In case the user doesn't specify the discriminator. I know it will then be able to match multiple people with same names, but maybe that's also the user's intention in such case.

(Also the same applies to `from:` filter)

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public static MessageFilter CreateFilter(string key, string value)
+        {
+            return key.ToLower() switch
+            {
+                "contains" => new ContainsMessageFilter(value),
+                "from" => new FromMessageFilter(value),
+                "has" => new HasMessageFilter(value),
+                "mentions" => new MentionsMessageFilter(value),
+                _ => throw new ArgumentException($"Invalid filter type '{key}'.", "key")
```
Tyrrrz, line 24:
```suggestion
                _ => throw new ArgumentException($"Invalid filter type '{key}'.", nameof(key))
```

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+    public abstract partial class MessageFilter
+    {
+        public abstract bool Filter(Message message);
+    }
+
+    public partial class MessageFilter
+    {
+        public static MessageFilter CreateFilter(string key, string value)
+        {
+            return key.ToLower() switch
```
Tyrrrz, line 18: Just to avoid any locale shenanigans

```suggestion
            return key.ToLowerInvariant() switch
```

Tyrrrz (commented):
Looking good so far! I haven't looked too deeply into the actual parsing logic/AST yet (need to get into a special mindset for that 😅), so I will do that at some point during the week.

DiscordChatExporter.Core/Exporting/Filtering/ContainsMessageFilter.cs
96-LB: Yeah, I think that's better. Discord search seems to mostly match the same way but also attempts to finds closely related words with some strange results ("AC" matches "ace", "youings" matches "you", etc.). Should we make this case-insensitive?

DiscordChatExporter.Core/Exporting/Filtering/MentionsMessageFilter.cs
96-LB: Sounds good.

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
96-LB: It looks like a `TargetInvocationException` with the message "Error: Exception has been thrown by the target of an invocation." I did some investigating and it seems that the reflection CliFx uses is wrapping the thrown exception because of the call to .Invoke(), with the original exception being stored as an InnerException. Setting an invalid partition produces the same behaviour.

DiscordChatExporter.Core/Exporting/Filtering/HasMessageFilter.cs
96-LB: Could we get away with just throwing a regex in here? `https?:\/\/\S\S` seems to match Discord's behaviour well enough (except for edge cases like "http://.." where the link body consists of only two of certain punctuation characters).

DiscordChatExporter.Core/Exporting/Filtering/FromMessageFilter.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        private readonly string _value;
+
+        public FromMessageFilter(string value) => _value = value;
+
+        public override bool Filter(Message message) =>
+            _value == message.Author.Name ||
+            _value == message.Author.FullName ||
+            _value == message.Author.Id.ToString();
+    }
```
Tyrrrz, line 15: For consistency:

```suggestion
            string.Equals(_value, message.Author.Name, StringComparison.OrdinalIgnoreCase) ||
            string.Equals(_value, message.Author.FullName, StringComparison.OrdinalIgnoreCase) ||
            string.Equals(_value, message.Author.Id.ToString(), StringComparison.OrdinalIgnoreCase);
    }
```

(may need `using System`)

DiscordChatExporter.Core/Exporting/Filtering/BooleanMessageFilter.cs
```diff
@@ -0,0 +1,27 @@
+﻿using DiscordChatExporter.Core.Discord.Data;
+
+namespace DiscordChatExporter.Core.Exporting.Filtering
+{
+    public class BooleanMessageFilter : MessageFilter
```
Tyrrrz, line 5: In parsing/formal language terminology, this type of syntax node is usually called a binary expression (i.e. 'equals' binary expression, 'and' binary expression, 'greater than' binary expression, etc). Maybe we can call it something like `BinaryExpressionMessageFilter`?

DiscordChatExporter.Core/Exporting/Filtering/BooleanMessageFilter.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        private readonly MessageFilter _first;
+        private readonly MessageFilter _second;
+        private readonly bool _unioned;
```
Tyrrrz, line 9: I'd probably use an enum field `_kind` for clarity:

```
public enum BinaryExpressionKind
{
    Or,
    And
}
```

To me, "unioned" doesn't really convey the right meaning.

DiscordChatExporter.Core/Exporting/Filtering/BooleanMessageFilter.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            _first = first;
+            _second = second;
+            _unioned = unioned;
+        }
+
+        public override bool Filter(Message message)
+        {
+            bool first = _first.Filter(message);
+            bool second = _second.Filter(message);
```
Tyrrrz, line 21: Can you please use `var` here and everywhere else where applicable?

```suggestion
            var first = _first.Filter(message);
            var second = _second.Filter(message);
```

DiscordChatExporter.Core/Exporting/Filtering/HasMessageFilter.cs
Tyrrrz: You can take this regex here (it's a bit more elaborate): https://github.com/Tyrrrz/DiscordChatExporter/blob/95cd6cb50cc1498a3512818c50f9990be0d01421/DiscordChatExporter.Core/Markdown/MarkdownParser.cs#L189-L192

Later we might figure out a way to unify it, but for now we can just copy-paste

DiscordChatExporter.Core/Exporting/Filtering/MentionsMessageFilter.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        private readonly string _value;
+
+        public MentionsMessageFilter(string value) => _value = value;
+
+        public override bool Filter(Message message) =>
+            message.MentionedUsers.Any(user =>
+                _value == user.Name ||
+                _value == user.FullName ||
+                _value == user.Id.ToString());
```
Tyrrrz, line 16: Let's use `string.Equals(...)` here with `StringComparison.OrdinalIgnoreCase` too.

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
Tyrrrz: Uh oh. Leave it like this for now, I will look at it separately. Thanks.

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+    public abstract partial class MessageFilter
+    {
+        public abstract bool Filter(Message message);
+    }
+
+    public partial class MessageFilter
+    {
+        public static MessageFilter CreateFilter(string text) => new ContainsMessageFilter(text);
+
+        public static MessageFilter CreateFilter(string key, string value)
```
Tyrrrz, line 17: These two probably don't need to be exposed?

```suggestion
        private static MessageFilter CreateFilter(string text) => new ContainsMessageFilter(text);

        private static MessageFilter CreateFilter(string key, string value)
```

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterTokenizer.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Superpower;
+using Superpower.Parsers;
+using Superpower.Tokenizers;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DiscordChatExporter.Core.Exporting.Filtering.Parsing
+{
+    public static class FilterTokenizer
```
Tyrrrz, line 10: What do you think, can we get away without having a tokenizer and just work with raw characters directly? I don't really care about error messages _too much_, but I wonder if writing a scanless parser would reduce complexity a bit.

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterParser.cs
```diff
@@ -0,0 +1,76 @@
...
+        public static TokenListParser<FilterToken, string> AnyString { get; } =
+            Token.EqualTo(FilterToken.QuotedString).Apply(FilterTextParsers.QuotedString)
+            .Or(Token.EqualTo(FilterToken.UnquotedString).Apply(FilterTextParsers.UnquotedString));
+
+        public static TokenListParser<FilterToken, MessageFilter> AnyFilter { get; } =
+            from minus in Token.EqualTo(FilterToken.Minus).Optional()
+            from content in KeyValuePair.Or(TextFilter.Try()).Or(GroupedFilter)
+            select minus.HasValue ? new NegatedMessageFilter(content) : content;
+
+        public static TokenListParser<FilterToken, MessageFilter> KeyValuePair { get; } =
```
Tyrrrz, line 23: Maybe `KeyValueFilter`?

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterParser.cs
```diff
@@ -0,0 +1,76 @@
...
+            select MessageFilter.CreateFilter(value);
+
+        public static TokenListParser<FilterToken, MessageFilter> MultipleFilter { get; } =
+            from first in AnyFilter
+            from rest in Token.EqualTo(FilterToken.VBar)
+                .IgnoreThen(AnyFilter)
+                .Select(value => (Value: value, Unioned: true)) //anonymous struct keeps track of whether a vertical bar was used
+                .Or(AnyFilter.Select(value => (Value: value, Unioned: false)))
+                .Many()
+            select rest.Aggregate(first, (current, next) => new BooleanMessageFilter(current, next.Value, next.Unioned));
```
Tyrrrz, line 40: Great use of `Aggregate`/left fold 🙂

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterParser.cs
```diff
@@ -0,0 +1,76 @@
...
+            from key in AnyString.Try()
+            from colon in Token.EqualTo(FilterToken.Colon).Try()
+            from value in AnyString
+            select MessageFilter.CreateFilter(key, value);
+
+        public static TokenListParser<FilterToken, MessageFilter> TextFilter { get; } =
+            from value in AnyString
+            select MessageFilter.CreateFilter(value);
+
+        public static TokenListParser<FilterToken, MessageFilter> MultipleFilter { get; } =
```
Tyrrrz, line 33: Maybe `BinarryExpressionFilter`?

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterParser.cs
```diff
@@ -0,0 +1,76 @@
...
+            from first in AnyFilter
+            from rest in Token.EqualTo(FilterToken.VBar)
+                .IgnoreThen(AnyFilter)
+                .Select(value => (Value: value, Unioned: true)) //anonymous struct keeps track of whether a vertical bar was used
+                .Or(AnyFilter.Select(value => (Value: value, Unioned: false)))
+                .Many()
```
Tyrrrz, line 39: I would do it a bit differently:

1. Have two separate parsers
2. One parser for || expression -> `OrBinaryExpressionFilter`
3. One parser for && expression -> `AndBinaryExpressionFilter`
4. Then have a higher-kind parser for both -> `OrBinaryExpressionFilter.Or(AndBinaryExpressionFilter).Many()` (or something like that)

Does it make sense?

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterParser.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public static TokenListParser<FilterToken, MessageFilter> GroupedFilter { get; } =
+            from open in Token.EqualTo(FilterToken.LParen)
+            from content in MultipleFilter
+            from close in Token.EqualTo(FilterToken.RParen)
+            select content;
+
+        public static TokenListParser<FilterToken, MessageFilter> Instance { get; } = MultipleFilter.AtEnd();
+
+        public static bool TryParse(string input, [MaybeNullWhen(false)] out MessageFilter value, [MaybeNullWhen(true)] out string error, out Position errorPosition)
```
Tyrrrz, line 50: Let's just do `public static MessageFilter Parse(string input)` and throw an exception directly in case of failure (instead of having to do it at the upper layers)

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterTextParsers.cs
```diff
@@ -0,0 +1,25 @@
...
+                .Many()
+            from close in Character.EqualTo('"')
+            select new string(content);
+
+        public static TextParser<string> UnquotedString { get; } =
+            from content in Character.EqualTo('\\').IgnoreThen(Character.In('"', '/')).Try()
+                .Or(Character.Except(c => char.IsWhiteSpace(c) || "():-|\"".Contains(c), "non-whitespace character except for (, ), :, -, |, and \""))
+                .AtLeastOnce()
+            select new string(content);
+    }
```
Tyrrrz, line 24: Can these also be inside `FilterParser`?

DiscordChatExporter.Core/Exporting/Filtering/ContainsMessageFilter.cs
Tyrrrz: Yeah let's make it case-insensitive but not bother with fuzzy matching (like ace, etc) for now

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterTokenizer.cs
96-LB: In terms of error messages I think the only real benefit of the tokenizer is being able to identify "string" since the rest are single characters anyway, and we could probably replicate that message elsewhere. I think being able to handle all whitespace in the tokenizer does reduce the parser complexity though, since that keeps us from having to worry about it in a bunch of other places. We probably wouldn't want to be rejecting things like `from: foo` vs. `from:foo`, and it might be messy to handle various cases like that in multiple parsers.

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterParser.cs
96-LB: For now I just entirely removed it and put the logic within the MessageFilter class since it uses both the tokenizer and parser. If we do remove the tokenizer, we can move it back if you'd like.

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterTextParsers.cs
96-LB: Done.

DiscordChatExporter.Core/Exporting/Filtering/BinaryExpressionMessageFilter.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        public override bool Filter(Message message)
+        {
+            var first = _first.Filter(message);
+            var second = _second.Filter(message);
+            return _kind == BinaryExpressionKind.Or ?
+                first || second :
+                first && second;
+        }
```
Tyrrrz, line 25: A bit more clear + short circuiting in case the second operand doesn't need to be evaluated.

```suggestion
        public override bool Filter(Message message) => _kind switch
        {
            BinaryExpressionKind.Or => _first.Filter(message) || _second.Filter(message),
            BinaryExpressionKind.And => _first.Filter(message) && _second.Filter(message), 
            _ => throw new InvalidOperationException($"Unknown binary expression kind '{_kind}'.")
        };
```

DiscordChatExporter.Core/Exporting/Filtering/HasMessageFilter.cs
```diff
@@ -0,0 +1,25 @@
...
+        public override bool Filter(Message message) =>
+            _value switch
+            {
+                "link" => Regex.IsMatch(message.Content, "https?://\\S*[^\\.,:;\"\'\\s]", DefaultRegexOptions),
+                "embed" => message.Embeds.Any(),
+                "file" => message.Attachments.Any(),
+                "video" => message.Attachments.Any(file => file.IsVideo),
+                "image" => message.Attachments.Any(file => file.IsImage),
+                "sound" => message.Attachments.Any(file => file.IsAudio),
+                _ => false
```
Tyrrrz, line 22: I forgot if we discussed this, but maybe it's better to throw here

```suggestion
                _ => throw new InvalidOperationException($"Invalid value provided for the 'has' message filter: '{_value}'");
```

DiscordChatExporter.Core/Exporting/Filtering/MessageFilter.cs
Tyrrrz: Yeah let's make them internal

DiscordChatExporter.Core/Exporting/Filtering/Parsing/FilterTokenizer.cs
Tyrrrz: Good point

Tyrrrz (commented):
Looks good! What do we have left?

DiscordChatExporter.Core/Exporting/Filtering/HasMessageFilter.cs
96-LB: I think we discussed whether `from:`/`mentions:` should throw with invalid input, but now that we accept usernames all input is valid for them. I agree that this one should throw; I'll make the change.

96-LB:
I think the last thing to worry about is adding an input to the GUI. I don't really have experience with WPF but I have a moment right now so I'll try to figure it out. There's also still the vague error whenever the input is invalid, but that's not stopping anything from working.

96-LB:
Hopefully that took care of it? Definitely worth a double check. I'm sure there are still small things to fix/improve, but I think this is mostly ready to go!

Tyrrrz:
Nice @96-LB!
Looks good. I'm ready to merge it. Do you want to test it a bit more or is it good to go?

96-LB:
I think it's good! It seems to be working as intended. If I missed anything, though, I'll submit an issue right away.

Tyrrrz:
Thanks! 🎉

PR #639, opened 2021-07-12
Update the in-app instructions for getting user token

### LINKED ISSUE #630: Update the in-app instructions for getting user token
### Version

2.28.0

### Flavor

### Export format

_No response_

### Details

It seems the old instructions for pulling a user token no longer work.  I [updated the wiki](https://github.com/Tyrrrz/DiscordChatExporter/wiki/Obtaining-Token-and-Channel-IDs#how-to-get-a-user-token) but could you please update the initial in-app instructions?

Perhaps it would be better to just link/load the wiki page, as these instructions can change again in the future.

Regarding the updated instructions, it seems you now need to enable the device toolbar (Device Mode) to get the token to appear.  Also, I don't believe CTRL+R does anything anymore but I left it in just in case.  Additionally, I couldn't update the GIF so you may want to update that or remove it.

### Steps to reproduce

- Open the app on a new/unconfigured profile/machine
- Old, no longer working instructions will appear

Yudi:
Closes #630

### DIFF
--- a/DiscordChatExporter.Gui/Views/RootView.xaml
+++ b/DiscordChatExporter.Gui/Views/RootView.xaml
@@ -1,4 +1,4 @@
-﻿<Window
+<Window
     x:Class="DiscordChatExporter.Gui.Views.RootView"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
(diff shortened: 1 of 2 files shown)


DiscordChatExporter.Gui/Views/RootView.xaml
```diff
@@ -163,23 +163,32 @@
...
+                            <Run Text="7. If the token key does not appear, press" />
                             <Run FontWeight="SemiBold" Text="Ctrl+R" />
                             <Run Text="to reload" />
                             <LineBreak />
-                            <Run Text="6. Find" />
-                            <Run FontWeight="SemiBold" Text="token" />
-                            <Run Text="at the bottom and copy the value" />
+                            <Run Text="8. Copy the" />
+                            <Run FontWeight="SemiBold" Text="value" />
+                            <Run Text="of the token key" />
```
Tyrrrz, line 191:
```suggestion
                            <Run Text="8. Copy the value of the" />
                            <Run FontWeight="SemiBold" Text="token" />
                            <Run Text="key" />
```

PR #644, opened 2021-07-13
Hide author icon when loading returns an error

### LINKED ISSUE #643: Text for Broken Embedded Author Icons overflows into embed contents
### Version

2.28

### Flavor

### Export format

HTML

### Details

When saving an embed, Author Icons with broken links currently show the broken file icon, with the Alt Text below it, running into the contents of the embed. Preferred behavior would be to hide the image, either with a blank space (As Discord does it) or by just hiding the element altogether.

This was also discussed in Issue #533.

### Steps to reproduce

(I used a starboard bot to do this, but it should be doable many different ways)
- Create an embed with an Author Icon linking to a user profile picture
- Break the Author Icon link (I believe a linked user changing their profile picture will do this, but I'm not 100% sure)
- The image will not appear in Discord's embed. Exporting through DCE now should display the broken file icon and alt text.

CliffJameston:
Closes #643


Tyrrrz:
Thanks!

PR #680, opened 2021-08-10
Removed reference to the total number of messages collected

Justman10000:
It's rather stupid, because you might want to share the generated files and nobody cares how many messages there are

Closes #ISSUE_NUMBER


96-LB:
Hi there! I happen to care about message counts.

CanePlayz:
Hi there! I happen to care about message counts.

Tyrrrz:
As already mentioned above, there are people that use this feature (it was actually removed at one point for unrelated reasons and users asked to bring it back). If you want, you can use your own fork that doesn't include the message count, but this won't be merged.

PR #802, opened 2022-02-05
Add support for stickers

### LINKED ISSUE #638: Support for stickers
### Flavor

_No response_

### Export format

HTML

### Details

Title says it all, support for the new custom and Discord-made stickers (as of now, messages with stickers are just displayed as blank messages in the HTML file).

Tyrrrz:
Closes #638

Blocking issues with lottie stickers:
  - When downloaded **without** media, sticker data cannot be retrieved due to CORS (Discord only serves it to `origin: discord.com`)
  - When downloaded **with** media, lottie library cannot retrieve the JSON from file system due to CORS (JS is not allowed to retrieve local content)

Thoughts on how to solve this?


Tyrrrz:
If you disable CORS through an extension it does work, for both PNG and LOTTIE backed stickers:

image

Issue #708, opened 2021-09-21
Long file paths crash the export process


DannyDorito24:
### Version

version 2.30.1

### Flavor

CLI (Command Line Interface)

### Details

When attempting to export a channel that contains media or links to media that are too long for the OS to handle, the export process hangs, then crashes. Maybe something could be added where if a filename is too long for the OS it truncates it to a certain length.

### Steps to reproduce

send a message in a channel such that it will embed an image with an extremely long filename, too long for your OS to handle
Export the channel with that message, and you will get "System.IO.PathTooLongException."

96-LB:
This is an OS limitation and varies by device. There's nothing really that can always be done to solve the problem. What if the export is run in a location that only allots 5 characters for media files? After using up a.png, b.png, etc., the exporter would run into trouble; meanwhile anything ending in `.jpeg` wouldn't even get a filename. Normally the exporter uses a hash to differentiate, but obviously in this case there's no room to insert one.

I'd imagine it's way too complicated to handle this exception in every case, but I do see how very long media filenames could be problematic and stop a channel from ever being exported. Maybe there could be a setting to limit the maximum length of included media files? That could at least stop a pathological (pun unintended) filename.

Tyrrrz:
> Maybe there could be a setting to limit the maximum length of included media files? That could at least stop a pathological (pun unintended) filename.

Currently we already truncate file names of downloaded media:

https://github.com/Tyrrrz/DiscordChatExporter/blob/ea31b1b270bfe77fdca6ad6eb5aecf98ae1b0780/DiscordChatExporter.Core/Exporting/MediaDownloader.cs#L112

(assuming that's what you meant)

96-LB:
Ah, yeah that is what I meant. 42 seems fairly reasonable, as you'd still need to be pretty deep in a filesystem to hit the 256-character path limit for Windows (I'm not familiar with the limit on other OS's though).

Tyrrrz:
I'm generally averse to automatically truncating paths based on OS max path length, because:

1. That would make the file names non-deterministic and unpredictable. Frankly, it leads to poor user experience if you run an export in one directory, get one set of file names, then run it another, and get another set. This is especially bad if the user specifically sets the output file path or template.

[...]

PR #812, opened 2022-02-17
Fix for extremely long file extensions

Tomlacko:
This fixes a bug where file names containing a dot with a long string after it would get wrongly split into a file name and a file extension, which could (due to file extensions not being truncated) cause a System.IO.PathTooLongException when the file gets created.
(The number 41 was chosen because 40 is one of the longest extensions used in practice that I was able to find, +1 for the dot.)

This was the main cause of issue #708 (I am friends with the OP and we worked on tracking it down), and while it is still technically possible to reach a file path that's too long for a given system, it should no longer happen by accident in normal use cases after this fix.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/MediaDownloader.cs
+++ b/DiscordChatExporter.Core/Exporting/MediaDownloader.cs
@@ -105,6 +105,7 @@ internal partial class MediaDownloader
         var fileExtension = Path.GetExtension(fileName);
         
         // Probably not a file extension, just a dot in a long file name
+        // https://github.com/Tyrrrz/DiscordChatExporter/issues/708
         if (fileExtension.Length > 41)
         {
             fileNameWithoutExtension = fileName;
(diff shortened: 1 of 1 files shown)


DiscordChatExporter.Core/Exporting/MediaDownloader.cs
```diff
@@ -103,7 +103,14 @@ private static string GetFileNameFromUrl(string url)
         // Otherwise, use the original file name but inject the hash in the middle
         var fileNameWithoutExtension = Path.GetFileNameWithoutExtension(fileName);
         var fileExtension = Path.GetExtension(fileName);
+        
+        // Probably not a file extension, just a dot in a long file name
```
Tyrrrz, line 107:
```suggestion
        // Probably not a file extension, just a dot in a long file name
        // https://github.com/Tyrrrz/DiscordChatExporter/issues/708
```

Tyrrrz:
Thanks!

PR #826, opened 2022-04-07
Use 512x512 user profile pictures

### LINKED ISSUE #655: Degraded image resolution of profile pictures
### Version

2.29.0

### Flavor

### Export format

HTML

### Details

Profile pictures get downloaded in a worse resolution than before (**40x40**), so low that you can hardly anything meaningful or faces.
This only affects when avatars are set, not the default Discord icon placeholder.
Before updating to 2.29.0 it would download them at **128x128** pixels.

Before:
cececd50fdc87b29929e65c768f24ad6
Now:
cececd50fdc87b29929e65c768f24ad6-9A0BE 

This doesn't seem to be a serverside issue, since the Discord program uses the normal 128 resolution for displaying avatars.

It is especially apparent when working with avatars where a lot is going on in the image.
And it makes it difficult or impossible to read text content in the avatar, leading to loss of information in the archival process.

Hoping this will be a fix without much issues, thanks.

### Steps to reproduce

- Download a channel with media enabled
- Look at the picture sizes of the user avatars

TheDude53:
It *may* be a bit overkill, but user profile pictures should be saved at a high quality for archival purposes.

Improves #685 and fixes #655 for good.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
+++ b/DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
@@ -293,12 +293,14 @@
             width: 40px;
             height: 40px;
             margin-right: 16px;
+            image-rendering: high-quality;
         }
 
         .chatlog__author-avatar {
             border-radius: 50%;
             width: 40px;
             height: 40px;
+            image-rendering: high-quality;
         }
 
         .chatlog__messages {
--- a/DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
+++ b/DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
@@ -293,7 +293,6 @@
             width: 40px;
             height: 40px;
             margin-right: 16px;
-            image-rendering: high-quality;
         }
 
         .chatlog__author-avatar {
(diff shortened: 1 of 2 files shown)


Tyrrrz:
Thanks for the PR.

If I remember correctly, last time the resolution of profile pictures was changed to a smaller one because there was a loss of quality when scaling them down to fit the profile picture dimensions in the HTML export (which is 40x40):

https://github.com/Tyrrrz/DiscordChatExporter/blob/bf1ad07a9cf0376e0717097cf7d60701bb686a56/DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml#L291-L302

Can you check if there is no quality loss?

TheDude53:
There doesn't appear to be any noticeable quality loss, however I've added some HQ scaling CSS just to stay on the safe side.

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
```diff
@@ -293,12 +293,14 @@
             width: 40px;
             height: 40px;
             margin-right: 16px;
+            image-rendering: high-quality;
```
Tyrrrz, line 296:
```suggestion
```

Tyrrrz:
Thanks, merged

PR #853, opened 2022-05-18
Add has:pin filter

### LINKED ISSUE #845: Export filter for pinned messages
### Flavor

_No response_

### Export format

_No response_

### Details

It would be nice to have a filter for pinned/previously pinned messages, maybe just "pinned:" or "has:pin" or something.

andrewkolos:
Closes #845.

Adds a new `has:*` filter, `has:pin`.
The filter filters for messages that are pinned.

Example: In a test DM channel, there is only one pinned message, with the text content `test`. There are many other non-pinned messages.

An export of the entire channel with the filter expression of `has:pin` results in the following output, which consists only of that one pinned message:

image

Please let me know if I understood the feature request correctly.


Tyrrrz:
Looks good, thanks a lot @andrewkolos!

PR #867, opened 2022-06-15
Update guide to direct users to the Advanced section of Discord settings rather than appearance

HanClinto:
Very small PR that is largely self-explanatory.  Developer Mode is now enabled via Advanced settings rather than in the Appearance tab, and I wanted to save other users the same confusion I had.

### DIFF
--- a/DiscordChatExporter.Cli/Commands/GuideCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/GuideCommand.cs
@@ -42,7 +42,7 @@ public class GuideCommand : ICommand
 
         console.Output.WriteLine(" 1. Open Discord");
         console.Output.WriteLine(" 2. Open Settings");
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Thanks! Can you also fix this in the CLI flavor too? It's in the `GuideCommand.cs`.

HanClinto:
I'm on OSX, so not able to test the GUI as easily.  That said, I did look through the source, and didn't find an equivalent section to this in [DashboardView.xaml](https://github.com/Tyrrrz/DiscordChatExporter/blob/master/DiscordChatExporter.Gui/Views/Components/DashboardView.xaml) (where the other help information seems to be).

Am I looking in the correct place?  Do we need to add a section here on how to enable developer mode?

Tyrrrz:
Never mind, you're right. This section only concerns CLI.

Issue #473, opened 2021-01-06
[FEATURE] Custom CSV output


cancerousname69:
The CSV output format currently lacks some information compared to the json output format. Adressing this by adding either custom CSV support or making the CSV format equivalent to the JSON format in data shown would be amazing.

96-LB:
The changelog contains the following text deep down in it (about a year ago):

> Please note that CSV export is most likely not going to receive future updates and will remain only as an option for those who wish to use it for text analysis. In other words, it won't contain any of the rich media features like embeds. If you're currently using CSV as a structured output to feed into your program, JSON output will replace that in the future.

The CSV export seems to be a predecessor to the JSON output and is now only intended for text features.

Tyrrrz:
Yeah, there's simply no good way to include the extra stuff in CSV, so it's not expanded anymore.

PR #868, opened 2022-06-15
Update CSV exporter to include embeds 

HanClinto:
Updates CSV exporter to include Embed information that was previously only accessible from JSON or plain text outputs. 
 However, CSV is generally much easier for me to parse, and I wanted to see some poll results had been submitted to the channel.
 
The text formatting of the embeds follows a similar convention as to how embeds are handled in the plain text exporter, and this method makes it only slightly more parseable through use of delimiters such as |, ;, and prefixes for embed properties (such as `Name: ` and `Field: `).

Example output of how complex embeds are exported from this script:

```
Name: >> cos1;Field: **Poll Question**=What is the condition of this Chain of Smog?;Field: ‚Äã=‚Äã;Field: **Anonymous**=False;Field: **Deadline**=Poll is closed.;Field: ‚Äã=**Final Results of the Poll (Single Choice):**
:regional_indicator_a: NM **- 1 Votes**
:regional_indicator_b: SP **- 4 Votes**
:regional_indicator_c: MP **- 12 Votes**
:regional_indicator_d: HP **- 9 Votes**
:regional_indicator_e: DMG **- 7 Votes**;Thumbnail: https://images-ext-2.discordapp.net/external/hjz95u_T6DWb9nVquxJ8WkLeMBhW-0FOA8dEP4PHhJ8/https/i.imgur.com/YksGRLN.png;Footer: React with ‚ùî to get info. It is not a vote option.;
```

Prior to this, it was impossible to see things like poll results without using a different output format.  It's not very pretty, but it's functional for my needs (and likely others who might want this functionality as well).

Note: The new column (`Embeds`) was added as the seventh and final column in the CSV so as to not break legacy dependencies that may expect the first six columns to remain unchanged in function and order.
 
Partially addresses the concerns of #473

### DIFF
--- a/DiscordChatExporter.Core/Exporting/Writers/CsvMessageWriter.cs
+++ b/DiscordChatExporter.Core/Exporting/Writers/CsvMessageWriter.cs
@@ -4,6 +4,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using DiscordChatExporter.Core.Discord.Data;
+using DiscordChatExporter.Core.Discord.Data.Embeds;
 using DiscordChatExporter.Core.Exporting.Writers.MarkdownVisitors;
 using DiscordChatExporter.Core.Utils.Extensions;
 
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Hi. Just like the linked issue explains, CSV format is not going to get feature updates in the future. JSON format is a much better format for programmable consumption and is easy to parse. Regardless of what language or platform you're using, parsing JSON should be trivial.

PR #869, opened 2022-06-17
Add filename token for current date

96-LB:
Adds the `%d` template token for the dynamic filename feature, which resolves to the curent date in `yyyy-MM-dd` format, a functionality which used to be possible using the `%b` token's default value but is no longer accessible.

This PR does not close any issues because it was discussed in the Discord. If you'd prefer, I can open one for posterity.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportRequest.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportRequest.cs
@@ -61,6 +61,7 @@ public partial record ExportRequest
                 "%P" => channel.Category.Position?.ToString() ?? "0",
                 "%a" => after?.ToDate().ToString("yyyy-MM-dd") ?? "",
                 "%b" => before?.ToDate().ToString("yyyy-MM-dd") ?? "",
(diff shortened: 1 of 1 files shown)


DiscordChatExporter.Core/Exporting/ExportRequest.cs
```diff
@@ -61,6 +62,7 @@ public partial record ExportRequest
                 "%P" => channel.Category.Position?.ToString() ?? "0",
                 "%a" => after?.ToDate().ToString("yyyy-MM-dd") ?? "",
                 "%b" => before?.ToDate().ToString("yyyy-MM-dd") ?? "",
+                "%d" => DateTime.Now.ToString("yyyy-MM-dd"),
```
Tyrrrz, line 65: I will replace this with `DateTimeOffset` for consistency (`SnowFlake.ToDate()` converts to `DateTimeOffset` too).

```suggestion
                "%d" => DateTimeOffset.Now.ToString("yyyy-MM-dd"),
```

PR #893, opened 2022-08-06
Print role details in JSON output

### LINKED ISSUE #892: Adding user role details into JSON exports
### Flavor

CLI (Command Line Interface)

### Export format

JSON

### Details

Hi guys, 

I really love this tool and have used it to dump multiple channels for some hobby analysis. However, I found user roles of message authors, which could be important information, missing in the JSON outputs. While role info has been crawled when retrieving user colors, it's really a waste not printing them out in detail in the final output.

I have prepared a PR for this feature, would be great if this could be added.

Adam

hahaschool:
Closes #892

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportContext.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportContext.cs
@@ -62,7 +62,13 @@ internal class ExportContext
             .Select(r => r.Color)
             .FirstOrDefault();
     }
+    public List<Role> TryGetMemberRoleList(Snowflake id)
+    {
+        var member = TryGetMember(id);
+        var roles = member?.RoleIds.Join(Roles, i => i, r => r.Id, (_, role) => role);
 
+        return roles?.ToList()??(new List<Role>());
+    }
     public async ValueTask<string> ResolveMediaUrlAsync(string url, CancellationToken cancellationToken = default)
     {
         if (!Request.ShouldDownloadMedia)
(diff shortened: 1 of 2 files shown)


Tyrrrz:
It doesn't build

hahaschool:
> It doesn't build

Seems to be a missing header, already added. I'm not a C# user so just feel free to make necessary amendments.

Thanks.

Adam

Tyrrrz:
I think it's better to have extended role information on each user object, to be consistent (we don't do lookups)

hahaschool:
> 

Do you mean for mentioned users we also print out all role details?

hahaschool:
If you mean it, then it's done. Please check.

Thanks.

Adam

Tyrrrz:
I mean here: https://github.com/Tyrrrz/DiscordChatExporter/blob/e8f8c47dbdaf32efe3e04aaf07eaa04bd2350a1f/DiscordChatExporter.Core/Exporting/Writers/JsonMessageWriter.cs#L274-L282

And here: https://github.com/Tyrrrz/DiscordChatExporter/blob/e8f8c47dbdaf32efe3e04aaf07eaa04bd2350a1f/DiscordChatExporter.Core/Exporting/Writers/JsonMessageWriter.cs#L211-L215

hahaschool:
Do you mean adding a field to User object? This may not be very reasonable since role is in the context of servers (i.e. same level as color), in case of DM there will be no role for a user. It's better to extend the Member object if we really want to do so, we can do this by:

1. We replace `roleIds` (list of IDs) to `roles` (list of Role objects), then we need to refactor user color implementation
2. We keep `roleIds` and add `roles` along with it, no breaks to user colors, but there will be some redundancy.

Otherwise we just keep the current setting aligning with the implementation of user color retrieval. 

Thanks.

Adam

Tyrrrz:
Hi Adam.

I decided against adding this because of dubious value in the context of exporting chats. DCE's JSON export is effectively a more structured clone of the HTML export, and it's not meant to be treated as a sort of general abstraction over Discord's API to dump arbitrary data. I recommend using Discord API directly for those cases. If you want, you can always continue using your fork, but I think this feature isn't going to add enough general value to justify the maintenance cost of adding it.

hahaschool:
> 

Hi Tyrrrz,

It's totally fine. Still, thank you for making this nice tool!

I will try to contribute again if I find anything useful to add.

Best,
Adam

PR #903, opened 2022-08-18
Expand tilde to home directory on *nix systems in output path

TSRBerry:
When specifying the output path like this `~/Documents/exportpath` new directories will be created in the current directory instead of using the correct path.

This PR fixes this issue by replacing "~" on non windows systems with the home directory for the output path.
I'm not quite happy with this solution yet, since I had to create a new private attribute and adapt the `OutputPath` property accordingly.
Also I think using `Replace()` for this operation might just be okay, but doesn't feel clean at all.
But it seems like .NET has no cross-platform pathlib library to handle this easily. Although a cleaner solution could be reached by using [this library](https://github.com/nemec/pathlib) for handling paths instead.

As a first time contributor I didn't want to include any new dependencies, so I chose to initially go with the first solution.

Let me know what you think and tell me if you want me to make further changes.

### DIFF
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -20,12 +20,21 @@ namespace DiscordChatExporter.Cli.Commands.Base;
 
 public abstract class ExportCommandBase : TokenCommandBase
 {
+    private string _outputPath = Directory.GetCurrentDirectory();
+
     [CommandOption(
         "output",
         'o',
         Description = "Output file or directory path."
     )]
-    public string OutputPath { get; init; } = Directory.GetCurrentDirectory();
+    public string OutputPath
+    {
+        get => _outputPath;
+        init
+        {
+            _outputPath = Path.GetFullPath(value);
+        }
+    }
 
     [CommandOption(
         "format",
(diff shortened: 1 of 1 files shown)


Tyrrrz:
I'm surprised .NET doesn't have anything built-in to handle it? Does `Path.GetFullPath(...)` or similar not work on tilde paths?

TSRBerry:
Initially I wanted to use this method too, but the documentation didn't mention anything about that, so I decided to do it differently.

But after seeing you asking about this too, I thought I could just test it real quick.

image

And sure enough, it works! Since this way cleaner than what I did, I'll replace my solution with this one.

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -20,12 +20,21 @@ namespace DiscordChatExporter.Cli.Commands.Base;
...
     )]
-    public string OutputPath { get; init; } = Directory.GetCurrentDirectory();
+    public string OutputPath
+    {
+        get => _outputPath;
+        init
+        {
+            _outputPath = Path.GetFullPath(value);
+        }
+    }
```
Tyrrrz, line 37:
```suggestion
    public string OutputPath
    {
        get => _outputPath;
        // Handle ~/ in paths on *nix systems
        // https://github.com/Tyrrrz/DiscordChatExporter/pull/903
        init => _outputPath = Path.GetFullPath(value);
    }
```

Tyrrrz:
Coolio, thanks :)

PR #920, opened 2022-09-09
Download emoji used in messages

### LINKED ISSUE #623: HTML --media option doesn't download emoji
### Version

v2.28.0

### Flavor

CLI (Command Line Interface)

### Export format

HTML

### Details

Using the HTML format with the `--media` option doesn't download emojis. They remain as links to Discord servers and `twemoji.maxcdn.com`. All emojis, especially custom ones, should be downloaded so that they're viewable offline or if the original Discord server is deleted.

### Steps to reproduce

1. Create a channel and post some messages containing single emojis.
2. Export it with the HTML format and the `--media` option.
3. Open the generated HTML file.

The HTML file contains lines such as
```html
<span class="preserve-whitespace"><img class="emoji emoji--large" alt="🎉" title="tada" src="https://twemoji.maxcdn.com/2/svg/1f389.svg"></span>
```
for Discord emojis and
```html
<span class="preserve-whitespace"><img class="emoji emoji--large" alt="foo" title="foo" src="https://cdn.discordapp.com/emojis/858851733346582541.png"></span>
```
for custom emojis. The files directory doesn't contain the emojis.

gan-of-culture:
Closes #623

As discussed in the issue -> added ResolveMediaUrlAsync to download the emoji image if wanted by the context. Currently emojis are only downloaded if used in for reactions, but with this change you can also download emojis used in messages.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/Writers/MarkdownVisitors/HtmlMarkdownVisitor.cs
+++ b/DiscordChatExporter.Core/Exporting/Writers/MarkdownVisitors/HtmlMarkdownVisitor.cs
@@ -123,7 +123,7 @@ internal partial class HtmlMarkdownVisitor : MarkdownVisitor
         var jumboClass = _isJumbo ? "chatlog__emoji--large" : "";
 
         _buffer
-            .Append($"<img loading=\"lazy\" class=\"chatlog__emoji {jumboClass}\" alt=\"{emoji.Name}\" title=\"{emoji.Code}\" src=\"{_context.ResolveMediaUrlAsync(emojiImageUrl)}\">");
+            .Append($"<img loading=\"lazy\" class=\"chatlog__emoji {jumboClass}\" alt=\"{emoji.Name}\" title=\"{emoji.Code}\" src=\"@await ResolveUrlAsync({emojiImageUrl})\">");
 
         return base.VisitEmoji(emoji);
     }
--- a/DiscordChatExporter.Core/Exporting/Writers/MarkdownVisitors/HtmlMarkdownVisitor.cs
+++ b/DiscordChatExporter.Core/Exporting/Writers/MarkdownVisitors/HtmlMarkdownVisitor.cs
@@ -123,7 +123,7 @@ internal partial class HtmlMarkdownVisitor : MarkdownVisitor
         var jumboClass = _isJumbo ? "chatlog__emoji--large" : "";
 
         _buffer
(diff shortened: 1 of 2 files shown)


gan-of-culture:
Needs testing. As soon as I got the results I will update the PR for review.

DiscordChatExporter.Core/Exporting/Writers/MarkdownVisitors/HtmlMarkdownVisitor.cs
```diff
@@ -123,7 +123,7 @@ protected override MarkdownNode VisitEmoji(EmojiNode emoji)
         var jumboClass = _isJumbo ? "chatlog__emoji--large" : "";
 
         _buffer
-            .Append($"<img loading=\"lazy\" class=\"chatlog__emoji {jumboClass}\" alt=\"{emoji.Name}\" title=\"{emoji.Code}\" src=\"{emojiImageUrl}\">");
+            .Append($"<img loading=\"lazy\" class=\"chatlog__emoji {jumboClass}\" alt=\"{emoji.Name}\" title=\"{emoji.Code}\" src=\"{_context.ResolveMediaUrlAsync(emojiImageUrl)}\">");
```
Tyrrrz, line 126: This needs to be awaited but we can't do it here

gan-of-culture: Hmm that makes it quite more difficult then since the message content is place into the HTML context using the @Raw(...) function. Is there an example I can maybe check out to resolve this in the best way?

Tyrrrz: Unfortunately no, I think it will take a bit of creativity to come up with a good way to solve this

gan-of-culture:
I know this solutions is very ugly and can break as soon as someone changes the emoji HTML template, but it works! Do you want to merge this?

(I don't have C# skills to come up with a better solution)

Tyrrrz:
We could probably just make the entire visitor class async so that your original solution works.

gan-of-culture:
I tried something like that but it just turned into a mess... so I think I'll use the regex implementation for personal use and close this PR and maybe someone can create a new PR later to fix it in a more correct way.

Maybe it would be smart to add your last comment to the issue as context on how to fix this correctly.

PR #926, opened 2022-09-12
Add HTML formatting for system messages

### LINKED ISSUE #571: "User pinned a message" not displayed correctly
### Flavor

### Details

If someone pins a message, it's shown like this in the HTML currently:

image

This is how it's normally shown in Discord:

image

### LINKED ISSUE #635: Show call messages properly
### Flavor

_No response_

### Export format

HTML

### Details

Currently, call messages are shown like this in the HTML file:

image

However, on Discord they are shown like this and should also be shown like that in the HTML file:

image

### LINKED ISSUE #641: Show join messages properly
### Flavor

_No response_

### Export format

HTML

### Details

I think it would be quite good to make the join messages look more like the ones in Discord, if that's possible.

How it's shown in the HTML file currently:
image

How it's shown on Discord:
image

gan-of-culture:
Closes #571
Closes #635 
Closes #641


gan-of-culture:
img

For testing purposes I started with adding pinned messages and calls. I introduced some new stuff to check if a message is a system message. There is new logic for system messages. System messages can only be grouped with other system messages. The link in the pinned message works as well.

Unfortunately the URLs to the icon .svg files can not be linked in the html file since the request will return a 403 HTTP error. So I was thinking about saving the .svg html tags and the render the on that is needed. 

Also would it be okay if I change the first letter of the "started call message content" to lower case since that is how discord displays it?

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
```diff
@@ -254,6 +254,22 @@
...
+        .chatlog__system-message-reference-link{
+            font-weight: 500;
+            color: #ffffff            
+        }
+
+        .chatlog__system-message-icon{
+            width: 16px;
+            height: 16px;
+            margin-right: 0.25rem;
+            border-radius: 50%;
```
Tyrrrz, line 270:
```suggestion
            color: #ffffff
        }

        .chatlog__system-message-icon{
            width: 16px;
            height: 16px;
            margin-right: 0.25rem;
            border-radius: 50%;
```

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
```diff
@@ -140,21 +190,24 @@
...
-                @if (!string.IsNullOrWhiteSpace(message.Content) || message.EditedTimestamp is not null)
+                @if (!isSystemMessage)
                 {
-                    <div class="chatlog__content chatlog__markdown">
-                        @{/* Text */}
-                        @if (!isContentHidden)
-                        {
-                            <span class="chatlog__markdown-preserve"><!--wmm:ignore-->@Raw(FormatMarkdown(message.Content))<!--/wmm:ignore--></span>
-                        }
+                    @if (!string.IsNullOrWhiteSpace(message.Content) || message.EditedTimestamp is not null)
```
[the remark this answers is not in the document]
gan-of-culture: the if statement would look something like this 
```
@if (!isSystemMessage && (!string.IsNullOrWhiteSpace(message.Content) || message.EditedTimestamp is not null))
```

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
gan-of-culture: Thanks for pointing it out. Removed the whitespaces^^

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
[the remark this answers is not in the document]
Tyrrrz: Okay, makes sense

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
```diff
@@ -62,7 +64,24 @@
...
+                        MessageKind.Call =>"https://discord.com/assets/7f75ab26bc5deb393cd8db44d9fe14e7.svg",
+                        MessageKind.ChannelNameChange => "",
+                        MessageKind.ChannelIconChange => "",
+                        MessageKind.ChannelPinnedMessage => "https://discord.com/assets/1f57f1c100434d1ff6569a495ada213e.svg",
+                        MessageKind.GuildMemberJoin => "",
+                        _ => ""
+                    };
+
+
+                    <img class="chatlog__system-message-icon" src="@await ResolveUrlAsync(iconUrl)" alt="System message icon" loading="lazy">
```
Tyrrrz, line 82:
```suggestion
                    };

                    <img class="chatlog__system-message-icon" src="@await ResolveUrlAsync(iconUrl)" alt="System message icon" loading="lazy">
```

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
```diff
@@ -81,7 +100,38 @@
...
+
+                    // system messages are grouped even if the message author is different
+                    // that's why we have to update the user values with the author of the current message
+                    userMember = Model.ExportContext.TryGetMember(message.Author.Id);
+
+                    userColor = Model.ExportContext.TryGetUserColor(message.Author.Id);
+
+                    userNick = message.Author.IsBot
+                        ? message.Author.Name
+                        : userMember?.Nick ?? message.Author.Name;
```
Tyrrrz, line 114: Can we just move the whole resolving logic from the top part to here, instead of overwriting the variables?

Tyrrrz:
Looks good, with a few comments. It's scary how complex the HTML template is becoming.

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
gan-of-culture: Hmm that thought crossed my mind too, but I thought to keep it the complexity low and readability high it makes more sense to have it at the top. (there is also a small performance gain I am pretty sure since most messages are not system messages)

The part in the loop is getting quite complex already so I only moved in the necessary parts.

gan-of-culture:
Closes #641

gan-of-culture:
how it is looking right now:
Join:
join

Call & Pin:
call

Can I make the first letter of the system messages lower case? Would that be something you're okay with?

Tyrrrz:
> Can I make the first letter of the system messages lower case? Would that be something you're okay with?

I assume to bring it in-line with Discord? Sure

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
```diff
@@ -787,6 +802,22 @@
...
+            </symbol>
+            <symbol id="icon-guildmemberjoin">
+                <title>GuildMemberJoin</title>
+                <g fill="none" fill-rule="evenodd"><path d="m18 0h-18v18h18z"/><path d="m0 8h14.2l-3.6-3.6 1.4-1.4 6 6-6 6-1.4-1.4 3.6-3.6h-14.2" fill="#3ba55c"/></g>
+            </symbol>
+        </defs>
+    </svg>
```
Tyrrrz, line 820: Can we merge this svg with the one above?

DiscordChatExporter.Core/Exporting/Writers/HtmlMessageWriter.cs
```diff
@@ -28,6 +28,13 @@ public HtmlMessageWriter(Stream stream, ExportContext context, string themeName)
     private bool CanJoinGroup(Message message) =>
         // First message in the group can always join
         _messageGroup.LastOrDefault() is not { } lastMessage ||
+        // Group system messages only with eachother
+        ( 
+            message.Kind.IsSystemMessage() &&
+            lastMessage.Kind.IsSystemMessage()
+        ) ||
+        // Join non system messages
+        !message.Kind.IsSystemMessage() &&
```
Tyrrrz, line 37: Can we simplify this somehow by moving the system message related checks to an additional check before it (with a return statement)? It's getting a little confusing

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
```diff
@@ -62,7 +64,11 @@
         <div class="chatlog__message">
             @{/* Left side */}
             <div class="chatlog__message-aside">
-                @if (isFirst)
+                @if (isSystemMessage)
+                {
+                    <svg class=chatlog__system-message-icon><use href="#icon-@Enum.GetName(typeof(MessageKind), message.Kind).ToLower()"></use></svg>
```
Tyrrrz, line 69: This should probably work too

```suggestion
                    <svg class=chatlog__system-message-icon><use href="#icon-@message.Kind.ToString().ToLowerInvariant()"></use></svg>
```

DiscordChatExporter.Core/Discord/Data/MessageKind.cs
```diff
@@ -12,4 +12,11 @@ public enum MessageKind
     ChannelPinnedMessage = 6,
     GuildMemberJoin = 7,
     Reply = 19
+}
+
+public static class MessageKindExtensions
+{
+    public static bool IsSystemMessage(this MessageKind c) => 
+        c != MessageKind.Default &&
+        c != MessageKind.Reply;
```
Tyrrrz, line 21:
```suggestion
    public static bool IsSystemMessage(this MessageKind c) => 
        c is not MessageKind.Default and not MessageKind.Reply;
```

gan-of-culture:
> generate the right text based on the message kind, separately for each format.

how would one do this exactly and where?^^ but yes that makes sense.

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
gan-of-culture: Yes it works! Just had to move it around a little bit.

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
gan-of-culture: Yes this also works totally fine. I just read online that using toString() is bad if you use it too often because it is slow. I didn't notice any noticeable slowdown so I think it fine :)

DiscordChatExporter.Core/Discord/Data/MessageKind.cs
gan-of-culture: This one is pretty straight forward.

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
```diff
@@ -81,7 +87,38 @@
...
+                    
+                        @{/* System message content */}
+                        @if(message.Kind == MessageKind.ChannelPinnedMessage)
+                        {
+                            <span class="chatlog__system-message"> pinned</span><span class="chatlog__system-message-reference-link chatlog__reference-link" onclick="scrollToMessage(event, @message.Reference.MessageId)"> a message</span><span class="chatlog__system-message"> to this channel.</span>
+                        }
+                        else
+                        {
+                            <span class="chatlog__system-message"> <!--wmm:ignore-->@Raw(FormatMarkdown(message.Content))<!--/wmm:ignore--></span>
+                        }
```
Tyrrrz, line 115: I mean here, do an `if`/`else` for each system message kind and generate the right content. Because, technically, system messages don't have content.

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
```diff
@@ -779,13 +794,30 @@
...
+            </symbol>
+            <symbol id="icon-guildmemberjoin" height="18" width="18">
+                <title>GuildMemberJoin</title>
+                <g fill="none" fill-rule="evenodd">
+                    <path d="m18 0h-18v18h18z"/>
+                    <path d="m0 8h14.2l-3.6-3.6 1.4-1.4 6 6-6 6-1.4-1.4 3.6-3.6h-14.2" fill="#3ba55c"/>
+                </g>
+            </symbol>
+        </defs>
```
Tyrrrz, line 820: Also can we make these consistent, i.e.:

- Follow the original naming convention of `xxx-icon` instead of `icon-xxx`
- Use `viewbox="0 0 <width> <height>"` instead of `width="..." height="..."`

DiscordChatExporter.Core/Exporting/Writers/HtmlMessageWriter.cs
```diff
@@ -37,6 +39,13 @@ public HtmlMessageWriter(Stream stream, ExportContext context, string themeName)
         // Other message must not be a reply
         message.Reference is null;
 
+    private bool CanJoinGroupSystemMessage(Message message) =>
+        // First message in the group can always join
+        _messageGroup.LastOrDefault() is not { } lastMessage ||
+        // Group system messages only with eachother
+        message.Kind.IsSystemMessage() &&
+        lastMessage.Kind.IsSystemMessage();
```
Tyrrrz, line 47: I mean more like:

```csharp
private bool CanJoinGroup(Message message)
{
	// First message in the group can always join
	if (_messageGroup.LastOrDefault() is not { } lastMessage)
  {
		return true;
	}

	// Group system messages with other system messages, regardless of author
	if (message.Kind.IsSystemMessage())
	{
		return lastMessage.Kind.IsSystemMessage();
	}

	// Group normal messages by author and time
	return
				// Must be from the same author
        lastMessage.Author.Id == message.Author.Id &&
        // Author's name must not have changed between messages
        string.Equals(lastMessage.Author.FullName, message.Author.FullName, StringComparison.Ordinal) &&
        // Duration between messages must be 7 minutes or less
        (message.Timestamp - lastMessage.Timestamp).Duration().TotalMinutes <= 7 &&
        // Other message must not be a reply
        message.Reference is null;
}
```

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
gan-of-culture: Changing name will create an issue with the <svg><use href=></use></svg> reference in the messagegrouptemplate.cshtml since the svg is referenced by the id which is "icon-" + message.Kind.ToString().ToLowerInvariant().

`<use href="#@message.Kind.ToString().ToLowerInvariant()-icon"` is this valid code?


I will adjust the viewbox - that's a good idea.

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
```diff
@@ -779,13 +794,30 @@
...
+    <svg style="display: none" xmlns="http://www.w3.org/2000/svg">
+        <defs>
+            <symbol id="attachment-icon" viewBox="0 0 720 960">
+                <path fill="#f4f5fb" d="M50,935a25,25,0,0,1-25-25V50A25,25,0,0,1,50,25H519.6L695,201.32V910a25,25,0,0,1-25,25Z"/>
+                <path fill="#7789c4" d="M509.21,50,670,211.63V910H50V50H509.21M530,0H50A50,50,0,0,0,0,50V910a50,50,0,0,0,50,50H670a50,50,0,0,0,50-50h0V191Z"/>
+                <path fill="#f4f5fb" d="M530,215a25,25,0,0,1-25-25V50a25,25,0,0,1,16.23-23.41L693.41,198.77A25,25,0,0,1,670,215Z"/>
+                <path fill="#7789c4" d="M530,70.71,649.29,190H530V70.71M530,0a50,50,0,0,0-50,50V190a50,50,0,0,0,50,50H670a50,50,0,0,0,50-50Z"/>
+            </symbol>
+            <symbol id="icon-channelpinnedmessage" fill="none" height="17" viewBox="0 0 17 17" width="17">
+                <title>ChannelPinnedMessage</title>
```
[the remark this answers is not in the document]
Tyrrrz: Yes please

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
Tyrrrz: Yes. Maybe you can make an extension method such that converts `ChannelPinnedMessage` to `Channel-Pinned-Message` (and then lowercase it). You can use RegEx to replace a pair of a lowercase character and an upper case character, with a dash in-between.

gan-of-culture: Please check how it looks now. Is this okay?

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
gan-of-culture: That is true, but is quite convenient actually I added some logic that makes the first letter of message.Content lowercase in the HTML export. Personally I think this is a better solution than a big `if`/`else`.

This is how it looks now:
new

DiscordChatExporter.Core/Exporting/Writers/Html/PreambleTemplate.cshtml
Tyrrrz: Yeah, this is great, thanks!

DiscordChatExporter.Core/Exporting/Writers/Html/MessageGroupTemplate.cshtml
Tyrrrz: I agree, I think it's a good compromise now

Tyrrrz:
Thanks!

PR #928, opened 2022-09-13
Fix scrollToMessage for pinned messages

gan-of-culture:
Fixes #926


gan-of-culture:
also fixes a bug where messages where falsely grouped with system messages

before the fix:
bfix

after the fix:
afix

Tyrrrz:
I already fixed it on master 🙃  Thank though

PR #947, opened 2022-10-08
Updated workflow

marcin-przywoski:
Updated workflow to publish trimmed packages

This will remove the unused assemblies while publishing hence reducing the size of release packages

Closes #ISSUE_NUMBER

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,10 +52,10 @@ jobs:
           dotnet-version: 6.0.x
 
       - name: Publish (CLI)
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Did you test if it works properly with all the reflection used in these projects?

marcin-przywoski:
Not yet - I'm already aware that it won't work with GUI so we can remove that line on GUI part. I wasn't aware the project is using reflection so I'll be doing some tests on that part

Tyrrrz:
Please test it first and then reopen the PR only if it works.

PR #952, opened 2022-10-15
Modify dockerfile to use non-root user

### LINKED ISSUE #851: Docker DCE CLI should not run as root by default on Linux machines
### Version

6621c10ae06b3f0fbf41915f8a9d470a6cccd0fd

### Flavor

CLI (Command Line Interface)

### Export format

_No response_

### Details

When running CLI via Docker, per wiki instructions, DCE runs as root. This is true as of both the `stable` build, and the latest `Dockerfile` available at HEAD. (6621c10ae06b3f0fbf41915f8a9d470a6cccd0fd as of this writing)

This seems like an unnecessary security compromise, and should be fixed to at least run as user 1000.

### Steps to reproduce

1.

```
docker run --volume /app/out:"$HOME/Discord"  tyrrrz/discordchatexporter export -c [is secret] -t [is also secret]
```

2.
```
sudo htop
```

3.

*Watch in awe as your security model is disregarded.*

BamButz:
Closes #851 

Modifies the docker image so that the CLI is not run by default as root.
Furthermore, the docker usage instructions (Wiki) should be modified so that all run commands
get the following argument `--user (id -u)` with this, the image will be started as the current user.

E.g. `docker run --rm --user $(id -u) -v /home/$(whoami):/out tyrrrz/discordchatexporter:latest`

### DIFF
--- a/DiscordChatExporter.Cli.dockerfile
+++ b/DiscordChatExporter.Cli.dockerfile
@@ -12,7 +12,9 @@ RUN dotnet publish DiscordChatExporter.Cli -c Release -o ./publish
 # Run
 FROM mcr.microsoft.com/dotnet/runtime:6.0 AS run
 
-COPY --from=build ./publish ./
+COPY --from=build ./publish /opt/dce
 
-WORKDIR ./out
-ENTRYPOINT ["dotnet", "/DiscordChatExporter.Cli.dll"]
+RUN useradd dce
+USER dce
+WORKDIR /out
+ENTRYPOINT ["dotnet", "/opt/dce/DiscordChatExporter.Cli.dll"]


DiscordChatExporter.Cli.dockerfile
```diff
@@ -12,7 +12,9 @@ RUN dotnet publish DiscordChatExporter.Cli -c Release -o ./publish
...
 
-COPY --from=build ./publish ./
+COPY --from=build ./publish /opt/dce
 
-WORKDIR ./out
-ENTRYPOINT ["dotnet", "/DiscordChatExporter.Cli.dll"]
\ No newline at end of file
+RUN useradd dce
+USER dce
+WORKDIR /out
```
Tyrrrz, line 19: For consistency. Is this right?

```suggestion
WORKDIR ./out
```

DiscordChatExporter.Cli.dockerfile
```diff
@@ -12,7 +12,9 @@ RUN dotnet publish DiscordChatExporter.Cli -c Release -o ./publish
 # Run
 FROM mcr.microsoft.com/dotnet/runtime:6.0 AS run
 
-COPY --from=build ./publish ./
+COPY --from=build ./publish /opt/dce
```
Tyrrrz, line 15: I'm not well-versed in nix systems, does copying it to `/opt/dce` make it available for the user `dce`? Could you please explain the solution to me like I'm 5 years old? 🙂

Tyrrrz (commented):
Thanks! Just a few questions

DiscordChatExporter.Cli.dockerfile
BamButz: In my opinion, is the usage of absolute paths a better practice and is unlikely to change unexpectedly.

DiscordChatExporter.Cli.dockerfile
BamButz: Actually, same reason as above. I think it is better practice to put the application files in a separate folder instead of the root.

It's like you're putting the application files straight on C:\ :)

PR #963, opened 2022-11-12
Add --throttle command line argument

slatinsky:
- Adds optional `--throttle <time_in_ms>` command line argument to delay each api call to Discord. 
- Delay is randomized time between <time_in_ms> and <time_in_ms>*2.
- Default is no throttle to keep backwards compatibility.

Example:
`--throttle 50` would delay each api call from 50 ms to 100 ms.

Why this change:
- The existing behaviour is to hit the API as frequently as it can and back off if it hits the api limit. This api limit would be frequently triggered on faster computers and internet connections closer to discord servers (low ping).
- This change would allow you to never hit api limit if you set high enough delay with tradeoff of slower export time.

Tested manually on command line version.
Closes no issue, because the project doesn't accept new feature requests.

### DIFF
--- a/DiscordChatExporter.Cli/Commands/Base/TokenCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/TokenCommandBase.cs
@@ -27,8 +27,14 @@ public abstract class TokenCommandBase : ICommand
     )]
     public bool IsBotToken { get; init; }
 
+    [CommandOption(
+        "throttle",
+        Description = "Delay each api request (in miliseconds)."
+    )]
+    public int Throttle { get; init; } = 0;
+
     private DiscordClient? _discordClient;
-    protected DiscordClient Discord => _discordClient ??= new DiscordClient(Token);
+    protected DiscordClient Discord => _discordClient ??= new DiscordClient(Token, Throttle);
 
     public abstract ValueTask ExecuteAsync(IConsole console);
 }
--- a/DiscordChatExporter.Core/Discord/DiscordClient.cs
+++ b/DiscordChatExporter.Core/Discord/DiscordClient.cs
@@ -20,11 +20,12 @@ namespace DiscordChatExporter.Core.Discord;
 public class DiscordClient
 {
     private readonly string _token;
+    private readonly int _throttle;
     private readonly Uri _baseUri = new("https://discord.com/api/v9/", UriKind.Absolute);
 
     private TokenKind? _resolvedTokenKind;
 
-    public DiscordClient(string token) => _token = token;
+    public DiscordClient(string token, int throttle) => (_token, _throttle) = (token, throttle);
 
     private async ValueTask<HttpResponseMessage> GetResponseAsync(
         string url,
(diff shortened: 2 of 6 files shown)


slatinsky:
Fixed GUI build and (manually) tested GUI version too

DiscordChatExporter.Cli/Commands/Base/TokenCommandBase.cs
```diff
@@ -27,8 +27,14 @@ public abstract class TokenCommandBase : ICommand
     )]
     public bool IsBotToken { get; init; }
 
+    [CommandOption(
+        "throttle",
+        Description = "Delay each api request (in miliseconds)."
+    )]
+    public int Throttle { get; init; } = 0;
```
Tyrrrz, line 34: Why not `TimeSpan`?

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -44,11 +45,18 @@ public class DiscordClient
...
+            var response = await Http.Client.SendAsync(
                 request,
                 HttpCompletionOption.ResponseHeadersRead,
                 innerCancellationToken
             );
+
+            if (_throttle != 0)
+            {
+                await Task.Delay(new Random().Next(_throttle, _throttle * 2), innerCancellationToken);
+            }
```
Tyrrrz, line 57: 1. Creating a `new Random()` defeats the purpose of using random. It should be `Random.Shared`
2. Why use random at all?
3. This should wait the throttle duration only if it hasn't already passed yet. I.e., you should subtract the time passed since the last request, because the work happening in between could have already introduced enough delay naturally

DiscordChatExporter.Cli/Commands/Base/TokenCommandBase.cs
Tyrrrz, line 34: Also, we should add this as an option to GUI as well

DiscordChatExporter.Core/Discord/DiscordClient.cs
slatinsky: Thanks for feedback

> Creating a new Random() defeats the purpose of using random. It should be Random.Shared

Random generator should be seeded from the system clock, but reusing Random instance is better according to documentation.

> 2\. Why use random at all?

Randomizing requests is common and very simple way to bypass bot detection

> Why not TimeSpan?

> This should wait the throttle duration only if it hasn't already passed yet. I.e., you should subtract the time passed since the last request

Because I used random, I didn't feel the need to perfectly time it.

But if we make randomization configurable (at least in CLI version), we should use TimeSpan.
For example `--throttle 50-100` would randomize requests between 50-100ms, but `--throttle 50` would send requests exactly after 50ms.

> Also, we should add this as an option to GUI as well

I agree. We should add it below "Parallel limit" in GUI->Global settings

slatinsky: > I wouldn't bother with that. If Discord wanted to know that you're using DCE (it already knows anyway), they wouldn't use request timing to detect that.

You are right. As seen with other (non discord) related open source scraping projects, they have teams to detect them and the only way to bypass it is to create custom private scraper. 

So we should drop randomization feature to make the code simpler.

> If your only goal with randomization is to avoid "bot detection" then I would drop it since it's not a viable masking method.

To avoid bot detection is not the (main) goal. The goal is to never hit api limit if you set high enough throttle delay.

slatinsky:
In the latest commit:
- Removed random
- Added "Delay requests (in ms)" to GUI (settings)
- Time passed since the last request is subtracted from throttle value

Tested manually both CLI and GUI

Tyrrrz:
Should be unnecessary now with the new rate limit handling, introduced in https://github.com/Tyrrrz/DiscordChatExporter/commit/49c4b12512fb3dd653a99b1808c82f590bc41a0d.

Issue #967, opened 2022-12-02
Docker unable to execute due to missing file


LIPDProductionsInc:
### Version

2.36.4

### Flavor

CLI (Command Line Interface)

### Details

In the latest version release, there appears to be no file named `DiscordChatExporter.Cli.dll` which cause docker to be unable to execute DiscordChatExporter on a Raspberry Pi.

### Steps to reproduce

- Follow the instructions in the [Docker Usage Instructions](https://github.com/Tyrrrz/DiscordChatExporter/wiki/Docker-usage-instructions)
- Execute `DiscordChatExporter`
- Receive Unhandled exception.

```
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly '/DiscordChatExporter.Cli.dll'. The system cannot find the file specified.

Filename: '/DiscordChatExporter.Cli.dll'
```

Tyrrrz:
Hi

> ...which cause docker to be unable to execute DiscordChatExporter on a Raspberry Pi

Are you pulling the image from the registry or building it yourself? I'm asking because, if I remember correctly, it took some extra steps to run the image on Raspberry Pi.

LIPDProductionsInc:
> Are you pulling the image from the registry or building it yourself? I'm asking because, if I remember correctly, it took some extra steps to run the image on Raspberry Pi.

I am following the steps in the Docker guide for the Raspberry Pi exactly as listed. Clone the git, CD into the new folder, and then build the image. I get the error on the last step

Tyrrrz:
Can you post as much of the build log as possible? I couldn't reproduce it

LIPDProductionsInc:
![image](https://user-images.githubusercontent.com/11858305/205423201-77e3dd78-2182-404e-822c-cb48342d01b4.png)
![image](https://user-images.githubusercontent.com/11858305/205423229-acbc8744-fdf6-4745-9d56-917062373d83.png)

LIPDProductionsInc:
Also the output of cloning the repo using `git clone http://github.com/Tyrrrz/DisscordChatExporter.git`
![image](https://user-images.githubusercontent.com/11858305/205423744-63fbd7f0-034f-4441-97aa-4948aa5e72de.png)

hyzchip:
I'm not entirely familiar with Docker, so this is just speculation...perhaps the issue has to do with the fact that the missing file is a DLL?  This appears to be a Linux system, which doesn't use DLLs unlike Windows.

[...]

Issue #879, opened 2022-06-28
Remove date format option


Tyrrrz:
### Details

Date format option was introduced a long time ago, where its usage was limited to message timestamps. It was introduced as a crude way to enable users to format timestamps in the export the same way they appear in their Discord client (in their respective locale).

Since then, a few things happened:

- We added support for embeds, which can contain timestamps in the footer. These timestamps are not formatted the same as other dates in the client.
- We improved the layout of the HTML export to show message time on hover. This timestamp only includes the time component and is, as a result, not formatted the same way as other dates in the client.
- We introduced the JSON export format, which formats all timestamps according to the way they are written when you call `JSON.stringify(...)` in JavaScript (which is a common practice in JSON outputs).

Reconciling these nuances with the existence of the date format option has become a nuisance. My suggestion is to remove the date format option and do the following instead:

- Format message timestamps according to user's system locale
- Format other timestamps the same way they are presented in Discord
- Additionally show tooltips with the timestamps formatted in a deterministic format that includes the time zone identifier

zabszk:
Tooltips are definitely better (and easier to use, no need to scroll to the preamble) than adding timezone to the preamble. Let's keep in mind that some people may want to use UTC timestamps (eg. #287). Currently it's possible by using the `u` date format, but it's suggested to remove it. So I think it's worth considering to use `u` format for the tooltips.

I'm not sure whether removing date format field completely is the best way. Maybe it would be better to just set the default value (on the first DCE launch) according to user's system locale? That way we don't remove the ability to customize the exports.

But overall I support this suggestion.

Tokariew:
What about… exporting messages to plain text… It's need some kind of time formatting…
`Format message timestamps according to user's system locale` And this option is not necessary nice, if you use docker… in which i don't think is option to setup easly locale…

zabszk:
> What about… exporting messages to plain text… It's need some kind of time formatting… `Format message timestamps according to user's system locale` And this option is not necessary nice, if you use docker… in which i don't think is option to setup easly locale…

That's why I think that instead of removing the date format option we should set the its default value according to the user's system locale (instead of setting it to some static default format). In that way we still allow user to customize the time format while making the default format friendly (so a lot of users don't need to customize it).

User may want to customize the format for a couple of reasons:
1. Exports can be made by multiple users from different countries (eg. within an organization) and they want to have the same format in all exports.
[...]

PR #974, opened 2022-12-27
Use Alpine as the base image in docker, publish a self-contained app

Zireael-N:
This reduces the image size from ~150 megabytes to ~80 megabytes, almost halving it. 

Also changed the order of `WORKDIR` and `COPY` directives in the resulting image to ensure that the build artifacts do not pollute (and do not conflict with existing files) the root directory. Added `WORKDIR` to the SDK image for the same reason.

Using relative paths in `WORKDIR` is actually against best practices - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#workdir - so I got rid of them too.

### DIFF
--- a/DiscordChatExporter.Cli.dockerfile
+++ b/DiscordChatExporter.Cli.dockerfile
@@ -18,15 +18,15 @@ RUN dotnet publish DiscordChatExporter.Cli \
 # Run
 FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine
 
-RUN adduser \
+RUN mkdir -p /opt/discord_chat_exporter && \
+    adduser \
     --disabled-password \
     --no-create-home \
     dce
 USER dce
+COPY --from=build /build/publish /opt/discord_chat_exporter
 
-WORKDIR /opt/discord_chat_exporter
-
-COPY --from=build /build/publish ./
+WORKDIR /out
 
 ENV PATH="$PATH:/opt/discord_chat_exporter"
 ENTRYPOINT ["DiscordChatExporter.Cli"]
--- a/DiscordChatExporter.Cli.dockerfile
+++ b/DiscordChatExporter.Cli.dockerfile
@@ -19,8 +19,8 @@ RUN dotnet publish DiscordChatExporter.Cli \
 FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine
 
 # tzdata is needed for DateTimeOffset.ToLocalTime (for TimeZoneInfo.Local, to be precise)
-RUN apk add --no-cache tzdata && \
-    adduser \
+RUN apk add --no-cache tzdata
+RUN adduser \
     --disabled-password \
     --no-create-home \
     dce
@@ -32,5 +32,7 @@ COPY --from=build /build/publish /opt/discord_chat_exporter
 # changing it would break existing workflows.
 WORKDIR /out
 
+# Having it in PATH is convenient for interactive shell sessions,
+# which may be useful for debugging.
 ENV PATH="$PATH:/opt/discord_chat_exporter"
 ENTRYPOINT ["DiscordChatExporter.Cli"]
--- a/DiscordChatExporter.Cli.dockerfile
+++ b/DiscordChatExporter.Cli.dockerfile
@@ -18,7 +18,8 @@ RUN dotnet publish DiscordChatExporter.Cli \
 # Run
 FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine
 
-RUN mkdir -p /opt/discord_chat_exporter && \
+RUN apk add --no-cache tzdata && \
+    mkdir -p /opt/discord_chat_exporter && \
     adduser \
     --disabled-password \
     --no-create-home \
--- a/DiscordChatExporter.Cli.dockerfile
+++ b/DiscordChatExporter.Cli.dockerfile
@@ -1,5 +1,7 @@
 # Build
-FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
+FROM mcr.microsoft.com/dotnet/sdk:7.0-alpine AS build
+
+WORKDIR /build
 
 COPY favicon.ico ./
 COPY NuGet.config ./
@@ -7,16 +9,24 @@ COPY Directory.Build.props ./
 COPY DiscordChatExporter.Core ./DiscordChatExporter.Core
 COPY DiscordChatExporter.Cli ./DiscordChatExporter.Cli
 
-RUN dotnet publish DiscordChatExporter.Cli --configuration Release --output ./publish
+RUN dotnet publish DiscordChatExporter.Cli \
+    --self-contained \
+    --use-current-runtime \
+    --configuration Release \
+    --output ./publish
 
 # Run
-FROM mcr.microsoft.com/dotnet/runtime:7.0 AS run
+FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine
 
-RUN useradd dce
+RUN adduser \
+    --disabled-password \
+    --no-create-home \
+    dce
 USER dce
 
-COPY --from=build ./publish ./
+WORKDIR /opt/discord_chat_exporter
 
-WORKDIR ./out
+COPY --from=build /build/publish ./
 
-ENTRYPOINT ["dotnet", "../DiscordChatExporter.Cli.dll"]
+ENV PATH="$PATH:/opt/discord_chat_exporter"
+ENTRYPOINT ["DiscordChatExporter.Cli"]
--- a/DiscordChatExporter.Cli.dockerfile
+++ b/DiscordChatExporter.Cli.dockerfile
@@ -27,6 +27,9 @@ RUN apk add --no-cache tzdata && \
 USER dce
 COPY --from=build /build/publish /opt/discord_chat_exporter
 
+# Need to keep this as /out for backwards compatibility with documentation.
+# A lot of people have this directory mounted in their scripts files, so
+# changing it would break existing workflows.
 WORKDIR /out
 
 ENV PATH="$PATH:/opt/discord_chat_exporter"
--- a/DiscordChatExporter.Cli.dockerfile
+++ b/DiscordChatExporter.Cli.dockerfile
@@ -18,8 +18,8 @@ RUN dotnet publish DiscordChatExporter.Cli \
 # Run
 FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine
 
+# tzdata is needed for DateTimeOffset.ToLocalTime (for TimeZoneInfo.Local, to be precise)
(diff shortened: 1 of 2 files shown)


Zireael-N:
Some of these changes might fix #967.

By the way, you might be interested in this - https://docs.docker.com/build/building/multi-platform/.

Tyrrrz:
Thanks a lot! This indeed looks way cleaner.

Tyrrrz:
One issue is that the documentation already suggests using `/out` as the mount path here: https://github.com/Tyrrrz/DiscordChatExporter/wiki/Docker-usage-instructions#usage

I know it might not be optimal, but can we avoid breaking changes?

Alternatively, is there a way to dynamically mount a path that corresponds to the current working directory in the container (in relation to `ENTRYPOINT`)? So that way we can avoid hard-codding it in documentation.

Zireael-N:
Changed the working directory back to `/out`.

> Alternatively, is there a way to dynamically mount a path that corresponds to the current working directory in the container (in relation to ENTRYPOINT)? So that way we can avoid hard-codding it in documentation.

You can capture the image's working directory like this - `DOCKER_PWD=$(docker run --entrypoint pwd --rm tyrrrz/discordchatexporter:stable)`. Looks pretty ugly to me though! I'd just add an explicit `--output` option to the example.

Zireael-N:
Another thing I should mention is that by default Alpine-based .NET images do not contain ICU and have `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` set to `true`. More on what's affected is here: https://github.com/dotnet/runtime/blob/8c58fc2347820ce48e09605d8adddb993df9ebb5/docs/design/features/globalization-invariant-mode.md.

If that's an issue, then I can add ICU to the image and override the environment variable. That bumps the image size to 121 megabytes. I didn't notice any usages of `CultureInfo.CurrentCulture` in the Core and CLI projects though.

Tyrrrz:
> You can capture the image's working directory like this - `DOCKER_PWD=$(docker run --entrypoint pwd --rm tyrrrz/discordchatexporter:stable)`. Looks pretty ugly to me though! I'd just add an explicit `--output` option to the example.

Yeah, that's ugly. Adding an explicit `--output` option, you mean with absolute path inside the container? I think it would ultimately result in the same issue, just in a different place, as we still would need to hard-code the path in the documentation.

> Another thing I should mention is that by default Alpine-based .NET images do not contain ICU and have `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` set to `true`. More on what's affected is here: [dotnet/runtime@`8c58fc2`/docs/design/features/globalization-invariant-mode.md](https://github.com/dotnet/runtime/blob/8c58fc2347820ce48e09605d8adddb993df9ebb5/docs/design/features/globalization-invariant-mode.md).
> 
> If that's an issue, then I can add ICU to the image and override the environment variable. That bumps the image size to 121 megabytes. I didn't notice any usages of `CultureInfo.CurrentCulture` in the Core and CLI projects though.

It shouldn't be an issue. At least not on purpose, but it'd be nice to test. We might need it eventually with #879 though.

DiscordChatExporter.Cli.dockerfile
```diff
@@ -1,22 +1,33 @@
...
 
-RUN useradd dce
+RUN apk add --no-cache tzdata && \
+    mkdir -p /opt/discord_chat_exporter && \
+    adduser \
+    --disabled-password \
+    --no-create-home \
+    dce
 USER dce
+COPY --from=build /build/publish /opt/discord_chat_exporter
```
Tyrrrz, line 28: Can we please split that `RUN` into separate commands for `apk add` and `adduser`? And ideally also add some comments above to remind future readers what they're needed for.

Also, is `mkdir` required here? I thought `COPY` creates directories.

DiscordChatExporter.Cli.dockerfile
```diff
@@ -1,22 +1,33 @@
...
 
-COPY --from=build ./publish ./
-
-WORKDIR ./out
+WORKDIR /out
```
Tyrrrz, line 30:
```suggestion
# Need to keep this as /out for backwards compatibility with documentation.
# A lot of people have this directory mounted in their scripts files, so
# changing it would break existing workflows.
WORKDIR /out
```

DiscordChatExporter.Cli.dockerfile
```diff
@@ -1,22 +1,33 @@
...
 
-ENTRYPOINT ["dotnet", "../DiscordChatExporter.Cli.dll"]
\ No newline at end of file
+ENV PATH="$PATH:/opt/discord_chat_exporter"
+ENTRYPOINT ["DiscordChatExporter.Cli"]
```
Tyrrrz, line 33: Can we just do `ENTRYPOINT ["/opt/discord_chat_exporter/DiscordChatExporter.Cli"]` here by the way? Or is there some extra benefit from having it on the path?

DiscordChatExporter.Cli.dockerfile
Zireael-N: > Also, is `mkdir` required here? I thought `COPY` creates directories.

You are absolutely correct, it's not needed.

> Can we please split that `RUN` into separate commands for `apk add` and `adduser`?

I can do that but usually people combine as many commands into a single `RUN` instruction as possible to minimize the amount of layers, see [Dockerfile @ golang](https://github.com/docker-library/golang/blob/7b9261231ef27db073c86c4fd7339b314703fe93/1.20-rc/alpine3.17/Dockerfile) or [Dockerfile @ zookeeper](https://github.com/31z4/zookeeper-docker/blob/5cf119d9c5d61024fdba66f7be707413513a8b0d/3.8.0/Dockerfile) for example.

DiscordChatExporter.Cli.dockerfile
Zireael-N: Yeah, you can use the full path. Having it in `$PATH` allows you to easily launch the app if you need to debug something while running an interactive shell session with `docker run --entrypoint /bin/sh -it --rm tyrrrz/discordchatexporter:stable`.

Zireael-N:
> Adding an explicit `--output` option, you mean with absolute path inside the container?

Yeah, that's what I meant. If it was there initially, changing the `WORKDIR` wouldn't be a breaking change.

> We might need it eventually with https://github.com/Tyrrrz/DiscordChatExporter/issues/879 though.

Leaving this for future reference then:

```diff
@@ -19,7 +19,11 @@
 FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine

 # tzdata is needed for DateTimeOffset.ToLocalTime (for TimeZoneInfo.Local, to be precise)
-RUN apk add --no-cache tzdata && \
+# icu-libs and icu-data-full are required for System.Globalization
+RUN apk add --no-cache \
+    tzdata \
+    icu-data-full \
+    icu-libs && \
     adduser \
     --disabled-password \
     --no-create-home \
@@ -32,5 +36,8 @@
 # changing it would break existing workflows.
 WORKDIR /out

-ENV PATH="$PATH:/opt/discord_chat_exporter"
+ENV PATH="$PATH:/opt/discord_chat_exporter" \
+    DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false \
+    LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8
 ENTRYPOINT ["DiscordChatExporter.Cli"]
```

DiscordChatExporter.Cli.dockerfile
Tyrrrz: > I can do that but usually people combine as many commands into a single `RUN` instruction as possible to minimize the amount of layers, see [Dockerfile @ golang](https://github.com/docker-library/golang/blob/7b9261231ef27db073c86c4fd7339b314703fe93/1.20-rc/alpine3.17/Dockerfile) or [Dockerfile @ zookeeper](https://github.com/31z4/zookeeper-docker/blob/5cf119d9c5d61024fdba66f7be707413513a8b0d/3.8.0/Dockerfile) for example.

Does it affect image size or only the build time? If it's just the latter, I think it's fine t

> I can do that but usually people combine as many commands into a single `RUN` instruction as possible to minimize the amount of layers, see [Dockerfile @ golang](https://github.com/docker-library/golang/blob/7b9261231ef27db073c86c4fd7339b314703fe93/1.20-rc/alpine3.17/Dockerfile) or [Dockerfile @ zookeeper](https://github.com/31z4/zookeeper-docker/blob/5cf119d9c5d61024fdba66f7be707413513a8b0d/3.8.0/Dockerfile) for example.

Does it affect the image size too or just the build time? If it's just the build time, I think it's fine to add a new layer since it's pretty fast as it is and I don't really build the image locally (I assume on CI the cache wouldn't matter anyway?).

DiscordChatExporter.Cli.dockerfile
Tyrrrz: I see, makes sense. Can you add a comment to explain that so I don't forget? :)

Zireael-N: Done!

DiscordChatExporter.Cli.dockerfile
Zireael-N: Doesn't seem to affect the image size, so I just went ahead and split them.

Zireael-N:
> It shouldn't be an issue. At least not on purpose, but it'd be nice to test. 

I tested it a bit and noticed a subtle change. `FileSize.ToString` uses the system's locale: https://github.com/Tyrrrz/DiscordChatExporter/blob/880f400e2c00b2372eae7e6ee0237f4efb25c778/DiscordChatExporter.Core/Discord/Data/Common/FileSize.cs#L42

If it's not intentional, you can use a new [string.Create overload introduced in .NET 6](https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/) to format it using `CultureInfo.InvariantCulture`:

```cs
string.Create(CultureInfo.InvariantCulture, $"{GetLargestWholeNumberValue():0.##} {GetLargestWholeNumberSymbol()}");
```

Tyrrrz:
> > It shouldn't be an issue. At least not on purpose, but it'd be nice to test.
> 
> I tested it a bit and noticed a subtle change. `FileSize.ToString` uses the system's locale:
> 
> https://github.com/Tyrrrz/DiscordChatExporter/blob/880f400e2c00b2372eae7e6ee0237f4efb25c778/DiscordChatExporter.Core/Discord/Data/Common/FileSize.cs#L42
> 
> If it's not intentional, you can use a new [string.Create overload introduced in .NET 6](https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/) to format it using `CultureInfo.InvariantCulture`:
> 
> ```cs
> string.Create(CultureInfo.InvariantCulture, $"{GetLargestWholeNumberValue():0.##} {GetLargestWholeNumberSymbol()}");
> ```

Indeed. Can you please include that in the PR?

Tyrrrz:
Thanks a lot!

PR #980, opened 2023-01-02
Remove regex check in `Snowflake.TryParse`

KubaZ2:
I think that the regex isn't needed here since the check can be done by setting right number styles.


Tyrrrz:
Thanks!

PR #985, opened 2023-01-10
Switch emoji CDN to jsDelivr

96-LB:
As a result of MaxCDN being shut down, our twemoji links stopped working (see https://github.com/twitter/twemoji/issues/580). This makes all standard emoji fail to render, and exports which download assets are painfully slow due to the CDN timing out every time we attempt to reach it. This PR resolves the issue by instead using jsDelivr to retrieve twemoji images.


Tyrrrz:
Thank you!

PR #986, opened 2023-01-15
Limit the number of messages gathered

carcabot:
Objective: To limit the number of messages gathered in the system to improve performance and reduce the risk of data overload.

This pull request implements a messages limit feature that will limit the number of messages gathered to a specified number set in command line. This will improve system performance by reducing the amount of data being processed and stored, and also reduce the risk of data overload.

Example:
`--limit 50` would limit the file generated to 50 messages


Tyrrrz:
Sorry, I have to close it as out of scope because the project is no longer accepting feature PRs.

PR #989, opened 2023-01-20
Add option to specify download directory for assets

### LINKED ISSUE #815: Allow specifying a common "assets directory" for exports that include downloading media
### Flavor

_No response_

### Export format

_No response_

### Details

Currently when downloading included media, all the files are downloaded into a directory with a name matching that of the exported file. That works great for single channel exports, but when exporting a guild or even just a few channels there's quite a bit of media that would be reused (e.g. avatars and the server icon). Coupled with the existing option to reuse existing media, by providing the option to speficy a "common directory" for these files we can save on both download bandwidth and storage space, which is great for large exports.

Bonus points if we can find a way to identify attachments/embedded media which have been duplicated in multiple channels, for embeds we could perhaps start by checking the URL from the referenced message, although for attachments the only way I can see would be to download and generate a file hash, then check for a match against an existing list of downloaded file hashes stored alongside the downloaded media. Not really sure if the additional time taken to compute all the file hashes would really outweigh the storage space reclaimed, especially since we'd need to download the file to hash it in the first place and gimmicks with Discord's reencoding would likely cause false negatives on images and videos. Avatars and the server icon should be fairly straightforward, though.

96-LB:
Closes #815 

Introduces a new option `--media-dir` to the CLI. When specified, all assets are downloaded to the specified directory, which can be formatted with template tokens. When not specified, the default behavior is the same as the current version (that is, save assets to a directory named `{filename}_Files`.

I didn't modify the code which transforms referenced the media path into a relative path, but that may not always be optimal behavior anymore. For example, if you dump all of your exports' assets into a common folder like `C:/DiscordChatExporterAssets`, using relative paths within your exports would actually break things if you want to move the files around later (which was the whole point of using relative paths in the first place). I figure that this problem is worth some discussion, so I didn't touch that in this PR.

### DIFF
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -106,6 +106,12 @@ public abstract class ExportCommandBase : TokenCommandBase
     )]
     public string DateFormat { get; init; } = "dd-MMM-yy hh:mm tt";
 
+    [CommandOption(
+        "fuck-russia",
+        Description = "Don't print the Support Ukraine message to the console."
+    )]
+    public bool IsUkraineSupportMessageDisabled { get; init; }
+
     private ChannelExporter? _channelExporter;
     protected ChannelExporter Exporter => _channelExporter ??= new ChannelExporter(Discord);
 
@@ -166,7 +172,8 @@ public abstract class ExportCommandBase : TokenCommandBase
                 {
                     try
                     {
-                        await progressContext.StartTaskAsync($"{channel.Category.Name} / {channel.Name}",
+                        await progressContext.StartTaskAsync(
+                            $"{channel.Category.Name} / {channel.Name}",
                             async progress =>
                             {
                                 var guild = await Discord.GetGuildAsync(channel.GuildId, innerCancellationToken);
@@ -257,12 +264,20 @@ public abstract class ExportCommandBase : TokenCommandBase
 
     public override ValueTask ExecuteAsync(IConsole console)
     {
-        // War in Ukraine message
-        console.Output.WriteLine("========================================================================");
-        console.Output.WriteLine("||   Ukraine is at war! Support my country in its fight for freedom   ||");
-        console.Output.WriteLine("||   Learn more: https://tyrrrz.me/ukraine                            ||");
-        console.Output.WriteLine("========================================================================");
-        console.Output.WriteLine("");
+        // Support Ukraine callout
+        if (!IsUkraineSupportMessageDisabled)
+        {
+            console.Output.WriteLine("┌────────────────────────────────────────────────────────────────────┐");
+            console.Output.WriteLine("│   Thank you for supporting Ukraine <3                              │");
+            console.Output.WriteLine("│                                                                    │");
+            console.Output.WriteLine("│   As Russia wages a genocidal war against my country,              │");
+            console.Output.WriteLine("│   I'm grateful to everyone who continues to                        │");
+            console.Output.WriteLine("│   stand with Ukraine in our fight for freedom.                     │");
+            console.Output.WriteLine("│                                                                    │");
+            console.Output.WriteLine("│   Learn more: https://tyrrrz.me/ukraine                            │");
+            console.Output.WriteLine("└────────────────────────────────────────────────────────────────────┘");
+            console.Output.WriteLine("");
+        }
 
         return default;
     }
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -127,7 +127,7 @@ public abstract class ExportCommandBase : TokenCommandBase
         }
 
         // Assets directory should only be specified when the download assets option is set
-        if (AssetsPath is not null && !ShouldDownloadAssets)
+        if (!string.IsNullOrWhiteSpace(AssetsPath) && !ShouldDownloadAssets)
         {
             throw new CommandException(
                 "Option --media-dir cannot be used without --media."
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -36,6 +36,20 @@ public abstract class ExportCommandBase : TokenCommandBase
         init => _outputPath = Path.GetFullPath(value);
     }
 
+    private readonly string? _assetsPath;
+
+    [CommandOption(
+        "media-dir",
+        Description = "Download assets to this directory."
+    )]
+    public string? AssetsPath
+    {
+        get => _assetsPath;
+        // Handle ~/ in paths on Unix systems
+        // https://github.com/Tyrrrz/DiscordChatExporter/pull/903
+        init => _assetsPath = value is not null ? Path.GetFullPath(value) : null;
+    }
+
     [CommandOption(
         "format",
         'f',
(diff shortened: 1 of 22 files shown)


DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -112,6 +126,14 @@ protected async ValueTask ExecuteAsync(IConsole console, IReadOnlyList<Channel>
             );
         }
 
+        // Assets directory should only be specified when the download assets option is set.
```
Tyrrrz, line 129: No period at the end of single-line comments (consistency)

```suggestion
        // Assets directory should only be specified when the download assets option is set
```

DiscordChatExporter.Core/Exporting/ExportRequest.cs
```diff
@@ -122,4 +139,20 @@ public partial record ExportRequest
...
+        Snowflake? before = null)
+    {
+        if (assetsPath is null)
+            return $"{basePath}_Files{Path.DirectorySeparatorChar}";
+
+        var actualAssetsPath = FormatPath(assetsPath, guild, channel, after, before);
+
+        return actualAssetsPath;
+    }
 }
```
Tyrrrz, line 158: Can we pull the `assetsPath is null` case outside in the property evaluator so that it becomes:

```csharp
public string OutputAssetsDirPath => _outputAssetsDirPath ??= (
    AssetsPath is not null
        ? GetOutputAssetsDirPath(
		        Guild,
		        Channel,
		        AssetsPath,
		        After,
		        Before
    		  )
        : $"{OutputBaseFilePath}_Files{Path.DirectorySeparatorChar}"
);
```

DiscordChatExporter.Core/Exporting/ExportRequest.cs
```diff
@@ -82,17 +91,15 @@ public partial record ExportRequest
         return PathEx.EscapeFileName(buffer.ToString());
     }
 
-    private static string GetOutputBaseFilePath(
+    private static string FormatPath(
```
Tyrrrz, line 94: Probably best to call it something like `EvaluateTemplateTokens`

Tyrrrz:
> I didn't modify the code which transforms referenced the media path into a relative path, but that may not always be optimal behavior anymore. For example, if you dump all of your exports' assets into a common folder like `C:/DiscordChatExporterAssets`, using relative paths within your exports would actually break things if you want to move the files around later (which was the whole point of using relative paths in the first place). I figure that this problem is worth some discussion, so I didn't touch that in this PR.

Oof, yeah. I'm not sure what's best in this case.

DiscordChatExporter.Core/Exporting/ExportRequest.cs
96-LB: Should we even bother with having a `GetOutputAssetsDirPath`? It'd be identical to `FormatPath`/`EvaluateTemplateTokens` with this change so we could probably just call the latter directly.

96-LB:
The three options I see are:
* Commit to always using relative/absolute paths, which would make moving certain exports painful depending on how you use the feature
* Use relative paths only when the default media path is being used and absolute paths if the user specifies a custom path, but this is probably a bad heuristic
* Add a toggle and let the user handle it, which satisfies every use case at the cost of more switches

I lean towards either the first or third, depending on how willing you are to add another command line option.

Tyrrrz:
> The three options I see are:
> 
> * Commit to always using relative/absolute paths, which would make moving certain exports painful depending on how you use the feature
> * Use relative paths only when the default media path is being used and absolute paths if the user specifies a custom path, but this is probably a bad heuristic
> * Add a toggle and let the user handle it, which satisfies every use case at the cost of more switches
> 
> I lean towards either the first or third, depending on how willing you are to add another command line option.

Do you personally use this feature? If so, how?

96-LB:
> Do you personally use this feature? If so, how?

I do daily backups into different folders (e.g. `.../backups/2023-01-21`,  `.../backups/2023-01-22`, etc), but I don't want to redownload identical assets daily so I have a shared media folder like `.../backups/media`. Sometimes I move individual exports around (I like to save monthly backups for longer in case channels/servers get deleted) so I think absolute paths work better for my purposes. However I also have no issue modding my own version to handle my personal use case.

Twi-Hard:
> However I also have no issue modding my own version to handle my personal use case.

I've really needed this as well.

Tyrrrz:
> > Do you personally use this feature? If so, how?
> 
> I do daily backups into different folders (e.g. `.../backups/2023-01-21`, `.../backups/2023-01-22`, etc), but I don't want to redownload identical assets daily so I have a shared media folder like `.../backups/media`. Sometimes I move individual exports around (I like to save monthly backups for longer in case channels/servers get deleted) so I think absolute paths work better for my purposes. However I also have no issue modding my own version to handle my personal use case.

I really don't want to add a switch for this, especially since this is specific to HTML exports. We can probably check if the assets path is outside the output directory (i.e. higher), then use the absolute path, and relative otherwise. But I'm not sure if it's a good heuristic, either.

In any case, I'm not sure that we should merge this if this implementation isn't very useful to people. As you mentioned, you'd rather use your own fork if the paths aren't absolute, and so far nobody else has provided their feedback.

96-LB:
> In any case, I'm not sure that we should merge this if this implementation isn't very useful to people. As you mentioned, you'd rather use your own fork if the paths aren't absolute, and so far nobody else has provided their feedback.

Understood. The linked issue seems fairly active, so maybe we can get clarification from some of the participants to see how they'd want this implemented.

For the sake of completeness, a fourth option that came to mind would be to simply let the user input a relative path if they want a relative path and an absolute path if they want an absolute path (e.g. `--media-dir "../../media"` vs `--media-dir "c:/backups/media"`). The advantage would be providing full flexibility without another command option, but the disadvantage would be that relative paths would be relative from the location of the output as opposed to the shell's current woring directory, which may be majorly unintuitive (and strips away the option to specify paths relative to the current directory).

CanePlayz:
> In any case, I'm not sure that we should merge this if this implementation isn't very useful to people. As you mentioned, you'd rather use your own fork if the paths aren't absolute, and so far nobody else has provided their feedback.

To provide my feedback: I really like this feature but I agree with you. An implementation that only supports relative or absolute paths probably shouldn't be merged.

Tyrrrz:
> Therefore, I second this idea. If I understood it correctly, it wouldn't work in 96-LB's case but in most others. If people using absolute paths decide to move their media folder or people using relative paths decide to move their media folder to another place or move the exports, I don't think there's much we can do anyways.

I'm not _entirely_ sure, but I think it should work for @96-LB's case too, no?

96-LB:
> I'm not *entirely* sure, but I think it should work for @96-LB's case too, no?

Yes, that heuristic would work for my case.

CanePlayz:
Whoops

96-LB:
@Tyrrrz Done. Let me know if you want anything else tweaked.

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -112,6 +126,14 @@ protected async ValueTask ExecuteAsync(IConsole console, IReadOnlyList<Channel>
             );
         }
 
+        // Assets directory should only be specified when the download assets option is set
+        if (AssetsPath is not null && !ShouldDownloadAssets)
```
Tyrrrz, line 130:
```suggestion
        if (!string.IsNullOrWhiteSpace(AssetsPath) && !ShouldDownloadAssets)
```

Tyrrrz:
Thanks!

PR #1032, opened 2023-05-12
Preliminary support for threads

### LINKED ISSUE #738: Show threads in CLI
### Flavor

CLI (Command Line Interface)

### Export format

_No response_

### Details

When running the channels command on a guild, making an option to also display all the threads in the guild would be cool. 

Currently in discord there is no way to list all threads in a server so adding this feature would make exporting threads 100x easier. Plus this would also allow for exporting all threads in a server much easier.

WAUthethird:
Since I have a fair bit of free time to spare, and since it seems that nobody else was going to do it, I've decided to work on a comprehensive threads/forums integration into DiscordChatExporter's GUI and CLI.

This first PR builds out the core of threads, and adds a proof of concept in the form of the CLI's `channels` command, which will now display a channel's threads below each channel. It is turned on by default, and can be turned off by using `--threads false`.

Full disclosure: I've actually never touched C# before this project. I'm primarily Python-oriented, but I've wanted this feature for a while, so I figured I'd dive in with this. This also means that I've likely implemented some things in a suboptimal or wrong manner, which I would appreciate feedback on before I proceed further. I'll give some specifics in a comment.

Closes #738.

### DIFF
--- a/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
@@ -59,9 +59,12 @@ public class GetChannelsCommand : DiscordCommandBase
 
                 foreach (var thread in threads)
                 {
+                    // Indent
+                    await console.Output.WriteAsync('\t');
+
                     // Thread ID
                     await console.Output.WriteAsync(
-                        thread.Id.ToString().PadLeft(26, ' ').PadRight(18, ' ')
+                        thread.Id.ToString().PadRight(18, ' ')
                     );
 
                     // Separator
--- a/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
@@ -20,6 +20,12 @@ public class GetChannelsCommand : DiscordCommandBase
     )]
     public required Snowflake GuildId { get; init; }
 
+    [CommandOption(
+        "threads",
+        Description = "Display threads alongside channels."
+    )]
+    public bool ThreadShow { get; init; } = true;
+
     public override async ValueTask ExecuteAsync(IConsole console)
     {
         var cancellationToken = console.RegisterCancellationHandler();
@@ -45,24 +51,27 @@ public class GetChannelsCommand : DiscordCommandBase
             using (console.WithForegroundColor(ConsoleColor.White))
                 await console.Output.WriteLineAsync($"{channel.Category.Name} / {channel.Name}");
 
-            var threads = (await Discord.GetGuildChannelThreadsAsync(channel.Id.ToString(), cancellationToken))
+            if (ThreadShow)
+            {
+                var threads = (await Discord.GetGuildChannelThreadsAsync(channel.Id.ToString(), cancellationToken))
                 .OrderBy(c => c.Name)
                 .ToArray();
 
-            foreach (var thread in threads)
-            {
-                // Thread ID
-                await console.Output.WriteAsync(
-                    thread.Id.ToString().PadLeft(25, ' ').PadRight(18, ' ')
-                );
+                foreach (var thread in threads)
+                {
+                    // Thread ID
+                    await console.Output.WriteAsync(
+                        thread.Id.ToString().PadLeft(26, ' ').PadRight(18, ' ')
+                    );
 
-                // Separator
-                using (console.WithForegroundColor(ConsoleColor.DarkGray))
-                    await console.Output.WriteAsync(" | ");
+                    // Separator
+                    using (console.WithForegroundColor(ConsoleColor.DarkGray))
+                        await console.Output.WriteAsync(" | ");
 
-                // Channel name / thread name
-                using (console.WithForegroundColor(ConsoleColor.White))
-                    await console.Output.WriteLineAsync($"{channel.Name} / {thread.Name}");
+                    // Channel name / thread name
+                    using (console.WithForegroundColor(ConsoleColor.White))
+                        await console.Output.WriteLineAsync($"Thread / {thread.Name}");
+                }
             }
         }
     }
--- a/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
@@ -54,8 +54,8 @@ public class GetChannelsCommand : DiscordCommandBase
             if (ThreadShow)
             {
                 var threads = (await Discord.GetGuildChannelThreadsAsync(channel.Id.ToString(), cancellationToken))
-                .OrderBy(c => c.Name)
-                .ToArray();
+                    .OrderBy(c => c.Name)
+                    .ToArray();
 
                 foreach (var thread in threads)
                 {
--- a/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
@@ -24,7 +24,7 @@ public class GetChannelsCommand : DiscordCommandBase
         "include-threads",
         Description = "Display threads alongside channels."
     )]
-    public bool ThreadShow { get; init; } = true;
+    public bool IncludeTHreads { get; init; }
 
     public override async ValueTask ExecuteAsync(IConsole console)
     {
@@ -51,7 +51,7 @@ public class GetChannelsCommand : DiscordCommandBase
             using (console.WithForegroundColor(ConsoleColor.White))
                 await console.Output.WriteLineAsync($"{channel.Category.Name} / {channel.Name}");
 
-            if (ThreadShow)
+            if (IncludeThreads)
             {
                 var threads = (await Discord.GetGuildChannelThreadsAsync(channel.Id.ToString(), cancellationToken))
                     .OrderBy(c => c.Name)
--- a/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
@@ -21,7 +21,7 @@ public class GetChannelsCommand : DiscordCommandBase
     public required Snowflake GuildId { get; init; }
 
     [CommandOption(
-        "threads",
+        "include-threads",
         Description = "Display threads alongside channels."
     )]
     public bool ThreadShow { get; init; } = true;
@@ -68,7 +68,7 @@ public class GetChannelsCommand : DiscordCommandBase
                     using (console.WithForegroundColor(ConsoleColor.DarkGray))
                         await console.Output.WriteAsync(" | ");
 
-                    // Channel name / thread name
+                    // Thread / thread name
                     using (console.WithForegroundColor(ConsoleColor.White))
                         await console.Output.WriteLineAsync($"Thread / {thread.Name}");
                 }
--- a/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
@@ -44,6 +44,26 @@ public class GetChannelsCommand : DiscordCommandBase
             // Channel category / name
             using (console.WithForegroundColor(ConsoleColor.White))
                 await console.Output.WriteLineAsync($"{channel.Category.Name} / {channel.Name}");
+
+            var threads = (await Discord.GetGuildChannelThreadsAsync(channel.Id.ToString(), cancellationToken))
+                .OrderBy(c => c.Name)
+                .ToArray();
+
+            foreach (var thread in threads)
+            {
+                // Thread ID
+                await console.Output.WriteAsync(
+                    thread.Id.ToString().PadLeft(25, ' ').PadRight(18, ' ')
+                );
+
+                // Separator
+                using (console.WithForegroundColor(ConsoleColor.DarkGray))
+                    await console.Output.WriteAsync(" | ");
+
+                // Channel name / thread name
+                using (console.WithForegroundColor(ConsoleColor.White))
+                    await console.Output.WriteLineAsync($"{channel.Name} / {thread.Name}");
+            }
         }
     }
 }
--- a/DiscordChatExporter.Core/Discord/Data/Thread.cs
+++ b/DiscordChatExporter.Core/Discord/Data/Thread.cs
@@ -1,0 +1,50 @@
+﻿using System.Linq;
+using System.Text.Json;
+using DiscordChatExporter.Core.Discord.Data.Common;
+using DiscordChatExporter.Core.Utils.Extensions;
+using JsonExtensions.Reading;
+
+namespace DiscordChatExporter.Core.Discord.Data;
+
+// https://discord.com/developers/docs/resources/channel#channel-object-example-thread-channel
+public partial record Thread(
+    Snowflake Id,
+    ChannelKind Kind,
+    Snowflake GuildId,
+    string Name,
+    Snowflake? LastMessageId) : IHasId
+{
+
+}
+
+public partial record Thread
+{
+    public static Thread Parse(JsonElement json)
+    {
+        var id = json.GetProperty("id").GetNonWhiteSpaceString().Pipe(Snowflake.Parse);
+        var kind = (ChannelKind)json.GetProperty("type").GetInt32();
+
+        var guildId =
+            json.GetPropertyOrNull("guild_id")?.GetNonWhiteSpaceStringOrNull()?.Pipe(Snowflake.Parse) ?? default;
+
+        var name =
+            // Guild channel
+            json.GetPropertyOrNull("name")?.GetNonWhiteSpaceStringOrNull() ??
+
+            // Fallback
+            id.ToString();
+
+        var lastMessageId = json
+            .GetPropertyOrNull("last_message_id")?
+            .GetNonWhiteSpaceStringOrNull()?
+            .Pipe(Snowflake.Parse);
+
+        return new Thread(
+            id,
+            kind,
+            guildId,
+            name,
+            lastMessageId
+        );
+    }
+}
--- a/DiscordChatExporter.Core/Discord/Data/ThreadChannel.cs
+++ b/DiscordChatExporter.Core/Discord/Data/ThreadChannel.cs
@@ -7,7 +7,7 @@ using JsonExtensions.Reading;
 namespace DiscordChatExporter.Core.Discord.Data;
 
 // https://discord.com/developers/docs/resources/channel#channel-object-example-thread-channel
-public partial record Thread(
+public partial record ThreadChannel(
     Snowflake Id,
     ChannelKind Kind,
     Snowflake GuildId,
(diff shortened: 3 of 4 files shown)


WAUthethird:
Two things I noticed that could need addressing:

1. My implementation of thread list retrieval is wrapped in a try/catch, as it currently tries to find threads in all channels. If the channel is inaccessible or is not a text/forum channel, it will throw an error. I'm not sure how to detect whether or not a channel is inaccessible beforehand, but it's not too difficult of a change to make to detect the type of channel. This makes the retrieval of threads pretty slow, which is probably the biggest issue at the moment.
2. The formatting of threads under channels in the CLI printout is likely not great, but I'd appreciate some feedback on it. Here's what it looks like:
Capture

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -200,6 +200,58 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
         return Guild.Parse(response);
     }
 
+    public async IAsyncEnumerable<Data.Thread> GetGuildChannelThreadsAsync(
+        string channelId,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        int currentOffset = 0;
+
+        var completeThreads = new List<JsonElement>();
```
Tyrrrz, line 209: Since this method returns `IAsyncEnumerable<...>`, you just yield elements asynchronously, instead of buffering them in a list (see further comments)

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -200,6 +200,58 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+
+        while (true)
+        {
+            var url = new UrlBuilder()
+                .SetPath($"channels/{channelId}/threads/search")
+                .SetQueryParameter("offset", currentOffset.ToString())
+                .Build();
+            try
+            {
+                var response = await GetJsonResponseAsync(url, cancellationToken);
```
Tyrrrz, line 219: You can use `TryGetJsonResponseAsync(...)` which returns `null` in case of an error status code. That way you don't have to `try`/`catch`.

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -200,6 +200,58 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+
+                completeThreads.AddRange(response.GetProperty("threads").EnumerateArray());
```
Tyrrrz, line 221: Here's how you should do it instead:

```suggestion
                foreach (var threadJson in response.GetProperty("threads").EnumerateArray())
                		yield return Thread.Parse(threadJson);
```

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -200,6 +200,58 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+                currentOffset += 25;
+            }
+            // If the channel is inaccessible, we cannot get the list of threads, which throws an error.
+            // I don't think there's any way to avoid this try/catch. Would love to be proven wrong.
+            catch (DiscordChatExporterException)
+            {
+                // There's probably a better way to mark response as empty?
+                var response = JsonDocument.Parse("{\"threads\": [], \"members\": [], \"total_results\": 0, \"has_more\": false}").RootElement;
+                completeThreads.AddRange(response.GetProperty("threads").EnumerateArray());
+                break;
```
Tyrrrz, line 238: Why is this necessary? Is just `break` not enough?

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -200,6 +200,58 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+            }
+        }
+
+        var threadArray = completeThreads.ToArray();
+
+        var responseOrdered = threadArray
+            .OrderBy(j => j.GetProperty("id").GetNonWhiteSpaceString().Pipe(Snowflake.Parse))
+            .ToArray();
```
Tyrrrz, line 246: Avoid ordering stuff at this level because it requires you to buffer data in memory. Let the client (i.e. DCE.CLI/DCE.GUI) do that if it needs to.

DiscordChatExporter.Core/Discord/Data/Thread.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Linq;
+using System.Text.Json;
+using DiscordChatExporter.Core.Discord.Data.Common;
+using DiscordChatExporter.Core.Utils.Extensions;
+using JsonExtensions.Reading;
+
+namespace DiscordChatExporter.Core.Discord.Data;
+
+// https://discord.com/developers/docs/resources/channel#channel-object-example-thread-channel
+public partial record Thread(
```
Tyrrrz, line 10: Probably should call it `ThreadChannel` to avoid naming ambiguity with `System.Threading.Thread`. That way you don't have to use `Data.Thread` everywhere.

Tyrrrz:
Does `channels/{channelId}/threads/search` return both active and archived threads?

What are the main advantages/disadvantages versus using these endpoints:

- `channels/{channel_id}/threads/archived/public`
- `channels/{channel_id}/threads/active`

WAUthethird:
> Does `channels/{channelId}/threads/search` return both active and archived threads?
> 
> What are the main advantages/disadvantages versus using these endpoints:
> 
>     * `channels/{channelId}/threads/archived/public`
> 
>     * `channels/{channelId}/threads/active`

Thanks for reviewing! I'll be going through your reviews and addressing each of them, but I wanted to do this one first.
Yes, `channels/{channelId}/threads/search` does return both active and archived threads.

`channels/{channelId}/threads/archived/public` seems to work okay for returning only archived threads, but I get `Only bots can use this endpoint` when using `channels/{channelId}/threads/active`, which is somewhat odd.

Since the current endpoint returns both active and archived, and the active endpoint you mention is bots-only for some reason, the current endpoint is the clear winner in my book.

DiscordChatExporter.Core/Discord/Data/Thread.cs
WAUthethird: Resolved in 92897399938aac018696b429597828cde5873238.

DiscordChatExporter.Core/Discord/DiscordClient.cs
WAUthethird: Resolved in f2c9bb1.

DiscordChatExporter.Core/Discord/DiscordClient.cs
WAUthethird: Resolved in 80cfb15.

DiscordChatExporter.Core/Discord/DiscordClient.cs
WAUthethird: Resolved in 80cfb15.

DiscordChatExporter.Core/Discord/DiscordClient.cs
WAUthethird: Resolved in 80cfb15.

DiscordChatExporter.Core/Discord/DiscordClient.cs
WAUthethird: Resolved in 80cfb15.

WAUthethird:
After implementing the changes you suggested, thread fetching is a lot faster!

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
```diff
@@ -20,6 +20,12 @@ public class GetChannelsCommand : DiscordCommandBase
     )]
     public required Snowflake GuildId { get; init; }
 
+    [CommandOption(
+        "threads",
+        Description = "Display threads alongside channels."
+    )]
+    public bool ThreadShow { get; init; } = true;
```
Tyrrrz, line 27: To keep it consistent with `export all` options, should we name this `--include-threads`/`IncludeThreads`?

https://github.com/Tyrrrz/DiscordChatExporter/blob/9048557b17ab7f51986458c30126260e9bd73761/DiscordChatExporter.Cli/Commands/ExportAllCommand.cs#L19-L29

DiscordChatExporter.Core/Discord/Data/ThreadChannel.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Linq;
+using System.Text.Json;
+using DiscordChatExporter.Core.Discord.Data.Common;
+using DiscordChatExporter.Core.Utils.Extensions;
+using JsonExtensions.Reading;
+
+namespace DiscordChatExporter.Core.Discord.Data;
+
+// https://discord.com/developers/docs/resources/channel#channel-object-example-thread-channel
+public partial record ThreadChannel(
```
Tyrrrz, line 10: I wonder if it would be easier to just use the existing `Channel` class for threads too. It has some extra stuff, but it should be nullable anyway. Might have to redesign the `ChannelCategory` concept though, since now it's more of a parent, which can either be a category (for channels) or a channel (for threads). Don't have to do it now, just thinking out loud.

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -200,6 +200,32 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+
+            var response = await TryGetJsonResponseAsync(url, cancellationToken);
+
+            foreach(var threadJson in response?.GetProperty("threads").EnumerateArray() ?? Enumerable.Empty<JsonElement>())
+                yield return ThreadChannel.Parse(threadJson);
+
+            if (!(response?.GetProperty("has_more").GetBoolean() ?? false))
+            {
+                break;
+            }
```
Tyrrrz, line 224: Can probably simplify further?

```suggestion
            var response = await TryGetJsonResponseAsync(url, cancellationToken);
            if (response is null)
            		break;

            foreach (var threadJson in response.GetProperty("threads").EnumerateArray())
                yield return ThreadChannel.Parse(threadJson);

            if (!response.GetProperty("has_more").GetBoolean())
                break;
```

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -200,6 +200,32 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+                currentOffset += 25;
```
Tyrrrz, line 225: We should probably not rely on a fixed number here and instead increment but the number of items that were returned in the search

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
WAUthethird: That seems like a good idea. I'll make that change.

DiscordChatExporter.Core/Discord/Data/ThreadChannel.cs
WAUthethird: Yeah, `ChannelCategory` is an interesting one. My thinking around how threads would be displayed in the GUI is pretty similar to categories - channels with threads get a dropdown for all of their threads, and threads can be selected/deselected as with channels.

There are a few issues with that approach, though:

- When selecting a channel for export, does it also select all threads belonging to it? If so, does the inverse apply when deselecting the channel?
- The same problem could occur when selecting individual/multiple threads. My thinking, though, is that thread selection should not affect the export status of the parent channel in any way.
- Because a channel would also be a dropdown, what sort of UI adjustment would need to happen in order to support selection/deselection of said parent channel in a non-confusing way? Currently, you just click the channel to mark it for export -- I suppose this could still work if we made the dropdown arrow behave like a small button?

Just some thoughts for the future, although perhaps these would be better suited for #999.

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
WAUthethird: Resolved in 62a58bb.

DiscordChatExporter.Core/Discord/DiscordClient.cs
WAUthethird: Wasn't able to get this one to compile due to this error:
`CS1061 'JsonElement?' does not contain a definition for 'GetProperty' and no accessible extension method 'GetProperty' accepting a first argument of type 'JsonElement?' could be found (are you missing a using directive or an assembly reference?)`

I got this error before, which I solved by using null-conditional and null-coalescing operators when parsing `response`. The same solution seems to work here.
Might be doing something wrong on my end? Your solution certainly seems to make more sense.

DiscordChatExporter.Core/Discord/DiscordClient.cs
WAUthethird: Resolved in 6183c87.

As with the thread parsing section, I got a few errors before I inserted the null-coalescing operator at the end:
```
CS0266 Cannot implicitly convert type 'int?' to 'int'. An explicit conversion exists (are you missing a cast?)
CS8629 Nullable value type may be null.
```
Still not sure if this sort of thing is the best thing to be doing, but it made the red line go away and the program seems to work fine.

DiscordChatExporter.Core/Discord/DiscordClient.cs
Tyrrrz: Ah, it has to be `response.Value.GetProperty("has_more")` because it's a value type

WAUthethird: Yep, you're right! That worked great.

Resolved in d88bbc8.

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
```diff
@@ -44,6 +50,29 @@ public override async ValueTask ExecuteAsync(IConsole console)
...
+                var threads = (await Discord.GetGuildChannelThreadsAsync(channel.Id.ToString(), cancellationToken))
+                .OrderBy(c => c.Name)
+                .ToArray();
+
+                foreach (var thread in threads)
+                {
+                    // Thread ID
+                    await console.Output.WriteAsync(
+                        thread.Id.ToString().PadLeft(26, ' ').PadRight(18, ' ')
+                    );
```
Tyrrrz, line 65: `PadRight(...)` before was used to ensure the `|` between columns was always in the same place, since channel IDs can actually have different length. Here it's better to just add indentation directly.

```suggestion
                    // Indent
                    await console.Output.WriteAsync('\t');

                    // Thread ID
                    await console.Output.WriteAsync(
                        thread.Id.ToString().PadRight(18, ' ')
                    );
```

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
```diff
@@ -44,6 +50,29 @@ public override async ValueTask ExecuteAsync(IConsole console)
             // Channel category / name
             using (console.WithForegroundColor(ConsoleColor.White))
                 await console.Output.WriteLineAsync($"{channel.Category.Name} / {channel.Name}");
+
+            if (ThreadShow)
+            {
+                var threads = (await Discord.GetGuildChannelThreadsAsync(channel.Id.ToString(), cancellationToken))
+                .OrderBy(c => c.Name)
+                .ToArray();
```
Tyrrrz, line 58: Consistency

```suggestion
                var threads = (await Discord.GetGuildChannelThreadsAsync(channel.Id.ToString(), cancellationToken))
                    .OrderBy(c => c.Name)
                    .ToArray();
```

Tyrrrz:
Looking good so far. We should definitely try to get it shown in the GUI. I'd say for now we should go with this:

- Expandable channels, just like with categories
- Threads have to be explicitly selected to get exported

The essence of expandable categories is here: https://github.com/Tyrrrz/DiscordChatExporter/blob/9048557b17ab7f51986458c30126260e9bd73761/DiscordChatExporter.Gui/Views/Components/DashboardView.xaml#L30-L32

And here: https://github.com/Tyrrrz/DiscordChatExporter/blob/9048557b17ab7f51986458c30126260e9bd73761/DiscordChatExporter.Gui/Views/Components/DashboardView.xaml#L311-L334

Hopefully, we should be able to add another layer to it with threads 🤔

WAUthethird:
> Looking good so far. We should definitely try to get it shown in the GUI. I'd say for now we should go with this:
> 
>     * Expandable channels, just like with categories
> 
>     * Threads have to be explicitly selected to get exported
> 
> 
> The essence of expandable categories is here:
> 
> https://github.com/Tyrrrz/DiscordChatExporter/blob/9048557b17ab7f51986458c30126260e9bd73761/DiscordChatExporter.Gui/Views/Components/DashboardView.xaml#L30-L32
> 
> And here:
> 
> https://github.com/Tyrrrz/DiscordChatExporter/blob/9048557b17ab7f51986458c30126260e9bd73761/DiscordChatExporter.Gui/Views/Components/DashboardView.xaml#L311-L334
> 
> Hopefully, we should be able to add another layer to it with threads 🤔

Yeah, for sure! I'll go with that. For now, I'm going to try to get threads fully integrated into the CLI (there's a few changes that need to happen in the export since the CLI technically already supports thread exports, but that's not explicit when looking at the export title or where it's saved), with full-server thread export with a few options for flexibility. I don't plan to go too deep into these options but it's mainly to ensure that everything works nicely in the core before embarking on the GUI changes.

WAUthethird:
I'll probably make those changes in another PR, so feel free to merge this one if it's good to go!

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
```diff
@@ -20,6 +20,12 @@ public class GetChannelsCommand : DiscordCommandBase
     )]
     public required Snowflake GuildId { get; init; }
 
+    [CommandOption(
+        "include-threads",
+        Description = "Display threads alongside channels."
+    )]
+    public bool ThreadShow { get; init; } = true;
```
Tyrrrz, line 27:
```suggestion
    public bool IncludeTHreads { get; init; }
```

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
```diff
@@ -44,6 +50,32 @@ public override async ValueTask ExecuteAsync(IConsole console)
             // Channel category / name
             using (console.WithForegroundColor(ConsoleColor.White))
                 await console.Output.WriteLineAsync($"{channel.Category.Name} / {channel.Name}");
+
+            if (ThreadShow)
```
Tyrrrz, line 54:
```suggestion
            if (IncludeThreads)
```

Tyrrrz:
Thanks!

tntmod54321:
Please please please include whether the thread is archived in the channel listing, for scheduled downloaders using a wrapper for DCE this would help a lot with wasting requests on threads that haven't been active for some time.

PR #1037, opened 2023-05-23
Use alternative endpoint to retrieve threads when requested by a bot

WAUthethird:
This PR refactors the recently added `GetChannelThreadsAsync` (now named `GetGuildThreadsAsync`) to return all threads in a single guild instead of all threads in a single channel. This means that the nonsensical API endpoints Discord created for bots will work in the same way as the single user endpoint already in place.

This PR does not address any specific Github issue, but was instead brought up as an issue in the Discord by @CanePlayz and @Tyrrrz.

### DIFF
--- a/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Linq;
+using System.Threading;
+using System.Threading.Channels;
 using System.Threading.Tasks;
 using CliFx.Attributes;
 using CliFx.Infrastructure;
@@ -36,6 +38,15 @@ public class GetChannelsCommand : DiscordCommandBase
             .ThenBy(c => c.Name)
             .ToArray();
 
+        var threads = Array.Empty<ChannelThread>();
+
+        if (IncludeThreads)
+        {
+            threads = (await Discord.GetGuildThreadsAsync(GuildId, cancellationToken))
+                .OrderBy(c => c.Name)
+                .ToArray();
+        }
+
         foreach (var channel in channels)
         {
             // Channel ID
@@ -51,13 +62,9 @@ public class GetChannelsCommand : DiscordCommandBase
             using (console.WithForegroundColor(ConsoleColor.White))
                 await console.Output.WriteLineAsync($"{channel.Category.Name} / {channel.Name}");
 
-            if (IncludeThreads)
+            foreach (var thread in threads)
             {
-                var threads = (await Discord.GetChannelThreadsAsync(channel.Id, cancellationToken))
-                    .OrderBy(c => c.Name)
-                    .ToArray();
-
-                foreach (var thread in threads)
+                if (thread.ParentId == channel.Id)
                 {
                     // Indent
                     await console.Output.WriteAsync('\t');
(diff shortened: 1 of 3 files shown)


CanePlayz:
Thanks

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
```diff
@@ -36,6 +36,15 @@ public override async ValueTask ExecuteAsync(IConsole console)
...
             .ToArray();
 
+        var threads = Array.Empty<ChannelThread>();
+
+        if (IncludeThreads)
+        {
+            threads = (await Discord.GetGuildThreadsAsync(GuildId, cancellationToken))
+                .OrderBy(c => c.Name)
+                .ToArray();
+        }
```
Tyrrrz, line 46: Can avoid mutation here:

```suggestion
        var threads = IncludeThreads
            ? (await Discord.GetGuildThreadsAsync(GuildId, cancellationToken))
                .OrderBy(c => c.Name)
                .ToArray()
            : Array.Empty<ChannelThread>();
```

DiscordChatExporter.Cli/Commands/GetChannelsCommand.cs
```diff
@@ -51,13 +60,9 @@ public override async ValueTask ExecuteAsync(IConsole console)
...
 
-            if (IncludeThreads)
+            foreach (var thread in threads)
             {
-                var threads = (await Discord.GetChannelThreadsAsync(channel.Id, cancellationToken))
-                    .OrderBy(c => c.Name)
-                    .ToArray();
-
-                foreach (var thread in threads)
+                if (thread.ParentId == channel.Id)
```
Tyrrrz, line 65: Can move this into the `foreach` with:

```csharp
foreach (var thread in threads.Where(t => t.ParentId == channel.Id)
```

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -311,30 +311,71 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
-                .SetPath($"channels/{channelId}/threads/search")
-                .SetQueryParameter("offset", currentOffset.ToString())
-                .Build();
-
-            var response = await TryGetJsonResponseAsync(url, cancellationToken);
-            if (response is null)
-                break;
-
-            foreach (var threadJson in response.Value.GetProperty("threads").EnumerateArray())
+            var response = await TryGetJsonResponseAsync($"guilds/{guildId}/threads/active", cancellationToken);
```
Tyrrrz, line 324: Can this actually return a non-200 response? Or would it just be an empty array in the response?

If the latter, then we should just use `GetJsonResponseAsync(...)` instead.

WAUthethird: Seems it's the latter, yeah. I tested a few different scenarios, and it doesn't seem like it'll ever return a non-200, except in cases where the guild is inaccessible by the bot. Resolved in d586bfd.

Tyrrrz:
Is this ready to be used now or do you need to make some more changes? (seeing as the PR is still in draft stage)

WAUthethird:
> Is this ready to be used now or do you need to make some more changes? (seeing as the PR is still in draft stage)

Sorry about that! I'll remember to next time.

Tyrrrz:
Cool! I'll review it in a bit.

Tyrrrz:
Thanks a lot, merged!

PR #1098, opened 2023-06-23
Add Interactive Export Command

mrkmg:
I know this project is in Maintenance mode, and I mostly wrote this just for my own use but figured I would offer it here.

This adds a simple CLI gui to select which channels to export.

Feel free to close if not interested in adding this new feature.


Tyrrrz:
Hey @mrkmg. Thanks for the PR. I understand its usefulness, but I don't want to bring interactive mode to the CLI. Sorry if this is not the conlusion you expected.

mrkmg:
No worries at all. Like I said was mostly my own use figured I would share.

PR #1101, opened 2023-07-01
Export role details in JSON format

slatinsky:
This pull request adds `roles` field to `messages.author` in JSON output format.

If no roles are available, empty array is created:
```json
"author": {
	...
	"roles": [],
	...
},
```
If role color is not available, `null` value is used. Else the color is is converted to hexcode

```json
"roles": [
	{
		"id": "712543702037017113",
		"name": "mod",
		"color": null,
		"position": 239
	},
	{
		"id": "766532373335259678",
		"name": "Orange",
		"color": "#F58231",
		"position": 218
	},
	...
]
```

Roles are sorted by position, as they are in the Discord client

Tested CLI export in Json format - DM channel (no roles) and Guild channel (with roles). \
GUI not tested, but the changes should not affect it.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportContext.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportContext.cs
@@ -106,6 +106,18 @@ internal class ExportContext
             .FirstOrDefault();
     }
 
+    public IEnumerable<Role> TryGetUserRoles(Snowflake id)
+    {
+        var member = TryGetMember(id);
+
+        return member?
+            .RoleIds
+            .Select(TryGetRole)
+            .WhereNotNull()
+            .OrderByDescending(r => r.Position)
+            .ToArray() ?? Array.Empty<Role>();
+    }
+
     public async ValueTask<string> ResolveAssetUrlAsync(string url, CancellationToken cancellationToken = default)
     {
         if (!Request.ShouldDownloadAssets)
--- a/DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
+++ b/DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
@@ -1,4 +1,5 @@
-﻿using System.IO;
+﻿using System.Collections.Generic;
+using System.IO;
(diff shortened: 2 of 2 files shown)


DiscordChatExporter.Core/Exporting/ExportContext.cs
```diff
@@ -106,6 +106,18 @@ await foreach (var role in Discord.GetGuildRolesAsync(Request.Guild.Id, cancella
             .FirstOrDefault();
     }
 
+    public IEnumerable<Role> TryGetUserRoles(Snowflake id)
```
Tyrrrz, line 109: Let's make this `IReadOnlyList<Role> GetUserRoles(...)` (no need for `Try...` because it doesn't return `null`).

Tyrrrz: Also, let's update `TryGetUserColor(...)` so that it calls `GetUserRoles(...)` for code reuse

slatinsky:
In the last commit:
- Renamed `IEnumerable<Role> TryGetUserRoles(...)` to `IReadOnlyList<Role> GetUserRoles(...)`
- Simplified `TryGetUserColor(...)`
- Tested build on Windows + CLI channel export in Json format

Tyrrrz:
Thanks!

stepech:
Hey, thanks for change.
Was there some consideration about how large the file will get with this change? Is it recorded really for each message including on servers with 100s of roles? What about creating extra file during export, which would hold just dictionary of users and their corresponding roles?

slatinsky:
Discord limits roles to 250 per server. That is the edge case scenario, but many users in my experience have only small number of roles. Yes, it is recorded in every message.
You could deduplicate a lot of things if you really wanted - reactions, authors, mentions, stickers. This felt like the best place to put it considering the current state of the project.
Yes, it will increase the file size. But text size increase is minimal in my opinion vs. the size of assets. And hard drive space is very cheap. And we were fetching this information anyway, just not recording it in the file.

PR #1102, opened 2023-07-06
Add export timestamp in JSON format

slatinsky:
This pull request adds a timestamp of time+date, when the JSON export was created. The format `yyyy-MM-ddTHH:mm:ss.ffzzz` is the same as seen with the other timestamps `timestamp`, `timestampEdited`,`callEndedTimestamp`,...

This is useful to know when the export was created to compare multiple exports of the same channel (for example profile pictures, nicknames change without modifying `timestampEdited` field in message object). And the feature has zero future maintenance cost.

Example - timestamp is added as `timestamp` field in the root object:
```json
{
  ...
  "dateRange": {
    "after": null,
    "before": null
  },
  "timestamp": "2023-07-06T17:02:25.02+00:00",
  "messages": [
    {
  ...
}
```

Tested CLI export in Json format - Guild channel.
GUI not tested, but the changes should not affect it.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
+++ b/DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
@@ -236,6 +236,9 @@ internal class JsonMessageWriter : MessageWriter
         _writer.WriteString("before", Context.Request.Before?.ToDate());
         _writer.WriteEndObject();
 
+        // Timestamp
+        _writer.WriteString("timestamp", System.DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.ffzzz"));
+
         // Message array (start)
         _writer.WriteStartArray("messages");
         await _writer.FlushAsync(cancellationToken);
--- a/DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
+++ b/DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
@@ -237,7 +237,7 @@ internal class JsonMessageWriter : MessageWriter
         _writer.WriteEndObject();
 
         // Timestamp
-        _writer.WriteString("timestamp", System.DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.ffzzz"));
+        _writer.WriteString("exportedAt", System.DateTimeOffset.UtcNow);
 
         // Message array (start)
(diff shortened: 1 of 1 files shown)


DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
```diff
@@ -236,6 +236,9 @@ await Context.ResolveAssetUrlAsync(Context.Request.Channel.IconUrl, cancellation
         _writer.WriteString("before", Context.Request.Before?.ToDate());
         _writer.WriteEndObject();
 
+        // Timestamp
+        _writer.WriteString("timestamp", System.DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.ffzzz"));
```
Tyrrrz, line 240: Can we use .NET's default JSON formatting of dates for this? It should be ISO 8601.

```suggestion
        _writer.WriteString("timestamp", DateTimeOffset.Now);
```

Tyrrrz: And maybe call this something `createdOn` to better convey what this timestamp relates to?

slatinsky: You are right about better naming. What about `exportedAt`? In the context of the channel export, `createdAt` may incorrectly suggest channel creation timestamp, not export timestamp.

Tyrrrz: Yeah, that's good

slatinsky:
`DateTimeOffset.Now` includes timezone (`+02:00` in my case):
```json
"exportedAt": "2023-07-09T00:38:32.1107474+02:00",
```

`DateTimeOffset.UtcNow` looks better:
```json
"exportedAt": "2023-07-08T22:44:06.3810376+00:00"
```

because other timestamps are in UTC (`+00:00`) too:
```json
"messages": [
    {
        ...
        "timestamp": "2023-06-19T21:29:07.354+00:00",
        "timestampEdited": null,
        ...
```

It doesn't matter in the end - time parser will parse both correctly - it will just look better if it is united.

Updated in the last commit.

Tested Build on Windows 10 .net 7.0 + tested CLI export in Json format.

Tyrrrz:
Thanks!

PR #1107, opened 2023-07-15
Export reaction users in JSON format

slatinsky:
This pull request adds users that reacted to a message with an emoji to JSON export

```json
"reactions": [
        {
          "emoji": {
            //...
          },
          "count": 37,
          "users": [
            {
              "id": "158600892784241132",
              "name": "user",
              "nickname": "user8745",
              "isBot": false,
              "avatarUrl": "..."
            },
            //...
```

- this change just adds a new `users` field - it doesn't break backwards compatibility
- user object in reactions is reduced - some fields are removed to improve export size (`JsonMessageWriter.cs:67`). This change doesn't affect users in other parts of export (backwards compatibility):
    - roles - for this we would need to request each user, too slow
    - color - requires roles
    - discriminator - not needed anymore
- only affecting JSON format - reaction users are fetched only if needed, so it doesn't affect other export formats other that json

notes:
- it affects only JSON format and doesn't break backwards compatibility. People exporting in JSON format expect this information to be included - I think that we don't need to add an option to turn this feature off

tested:
- pagination (> 100 users per emoji)
- windows 10 build + CLI export to JSON format


DiscordChatExporter.Core/Exporting/ExportContext.cs
```diff
@@ -138,4 +138,10 @@ public async ValueTask<string> ResolveAssetUrlAsync(string url, CancellationToke
             return url;
         }
     }
+
+    internal async Task<IEnumerable<User>> ResolveReactionUsers(Snowflake messageId, Emoji emoji, CancellationToken cancellationToken)
+    {
+        var users = await Discord.GetMessageReactionsAsync(Request.Channel.Id, messageId, emoji, cancellationToken);
+        return users;
+    }
```
Tyrrrz, line 146: Since this doesn't rely on any state, you could just call `Context.Discord.GetMessageReactionsAsync(...)` instead, and remove this method.

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -470,4 +470,45 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+        Snowflake messageId,
+        Emoji emoji,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+
+        var reactionName = emoji.Name;  // default emoji
+        if (emoji.Id.HasValue) {        // custom emoji
+            reactionName = reactionName + ":" + emoji.Id.Value.ToString();
+        }
+        reactionName = Uri.EscapeDataString(reactionName);
```
Tyrrrz, line 485:
```suggestion
        var reactionName = Uri.EscapeDataString(
            emoji.Id is not null
                // Custom emoji
                ? emoji.Name + ':' + emoji.Id
                // Standard emoji
                : emoji.Name
        );
```

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -470,4 +470,45 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+
+        var currentAfter = Snowflake.Zero;
+
+        while (true) {
+            var url = new UrlBuilder()
```
Tyrrrz, line 490:
```suggestion
        while (true)
        {
            var url = new UrlBuilder()
```

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -470,4 +470,45 @@ private async ValueTask<TokenKind> GetTokenKindAsync(CancellationToken cancellat
...
+
+            foreach (var user in newUsers) {
+                yield return user;
+            }
+
+            if (newUsers.Length < 100) {
+                yield break;
+            }
+
+            currentAfter = newUsers.Last().Id;
```
Tyrrrz, line 511:
```suggestion
            var users = response.EnumerateArray().Select(User.Parse).ToArray();

            if (!users.Any())
                yield break;

            foreach (var user in users)
            {
                yield return user;
                currentAfter = user.Id;
            }

            // Each batch can contain up to 100 users.
            // If we got fewer, then it's definitely the last batch.
            if (users.Length < 100)
                yield break;
```

DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
```diff
@@ -64,6 +64,31 @@ public JsonMessageWriter(Stream stream, ExportContext context)
...
+            "avatarUrl",
+            await Context.ResolveAssetUrlAsync(
+                Context.TryGetMember(user.Id)?.AvatarUrl ?? user.AvatarUrl,
+                cancellationToken
+            )
+        );
+
+        _writer.WriteEndObject();
+        await _writer.FlushAsync(cancellationToken);
+    }
```
Tyrrrz, line 90: If this method is only used in one place, then you can just inline it. Make sure to add a comment explaining that it only writes a limited version of the user.

DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
```diff
@@ -360,6 +385,14 @@ await Context.ResolveAssetUrlAsync(Context.Request.Channel.IconUrl, cancellation
 
             _writer.WriteNumber("count", reaction.Count);
 
+            _writer.WriteStartArray("users");
+
+            var users = await Context.ResolveReactionUsers(message.Id, reaction.Emoji, cancellationToken: cancellationToken);
+            foreach (var user in users)
+                await WriteReactionUserAsync(user, cancellationToken);
```
Tyrrrz, line 392: You can do:

```csharp
await foreach (var user in Context.Discord.GetMessageReactionsAsync(...))
```

To do this with minimal buffering.

slatinsky:
Changes:
- rebased with upstream
- readded discriminator, because some users are still not migrated
- refactored according to the code review

Tested on CLI build Windows 10, export to JSON format. Pagination (> 100 users per emoji) works too

Tyrrrz:
Thanks, merged. I'll clean up in a follow up commit.

PR #1111, opened 2023-07-26
removed shitty politics

iloveyumyumshrimpnoodles:
Keep politics out of code, you think i want to see a message saying "ooh muhh ukraine in danger, slava ukraine pls else don't use >:(" when i want (or somebody else wants) to export discord chats? You think that does something?

If you want to do something about the war, put on a soldiers uniform and grab a gun and fight.

Fuckoff

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -182,7 +182,6 @@
 
 ## v2.33 (06-Mar-2022)
 
-- Added messages informing about war in Ukraine and available ways to help.
 - Added support for rendering stickers in HTML and JSON. Lottie-based stickers currently cannot be displayed in HTML exports (see [#803](https://github.com/Tyrrrz/DiscordChatExporter/issues/803)).
 - Added a new `reaction:` message filter which can be used to check if someone reacted to a message with a specific emoji. You can either pass the emoji name (e.g. `reaction:LUL`) or its ID (e.g. `reaction:41771983429993937`).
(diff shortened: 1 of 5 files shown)


MyFedora:
Perhaps try to be more constructive about it, @Tyrrrz?

I believe the following actions are a bad look for any maintainer really:
- [Deleting issues because you disagree with their opinion](
https://web.archive.org/web/20230726172642/https://github.com/Tyrrrz/DiscordChatExporter/issues/979).
- Editing other people's comments because you disagree with their opinion.
- Deleting the edit history so that others have to go to the waybackmachine to see the real comment.

It's frankly the textbook definition of censorship.

Let everyone dump complaints related to the Ukraine support messages in an issue and ignore it, so that you don't have to deal with that in other communication channels.

I _personally_ believe that such messages are unnecessary. Everyone is aware of the Ukraine Russia war already. If they didn't donate when the war hit the news, they won't donate when a software bugs them about it. And I _personally_ don't donate to any country waging a war as every participant forces and/or tries to manipulate their population to go die for their fatherland instead of escaping with their friends and family. A country isn't worth sacrificing your life for.

Technetium1:
@MyFedora Edit history can't be deleted _by maintainers_, at least on GitHub. If something is deleted, it's probably because it was reported and removed by GitHub staff. Normally this behavior would be frowned upon, but it's not really inappropriate here. There are constant spam issues and PRs, justifying increasingly strong measures. Nobody seems to get the concept of just **forking off** if they don't like it! 🙃

PR #1125, opened 2023-08-31
Optimize fetching threads

slatinsky:
By adding new threads support, `Fetching channels...` stage takes too long for bigger guilds with many threads. Some data doesn't need to be fetched, because we know, that those channels/threads won't be exported.

If `--after` parameter is specified, the main speed-up is to break early of thread fetching, by sorting threads by `LastMessageId` (endpoint `channels/{channel.Id}/threads/search` - query parameters `sort_by=last_message_time&sort_order=desc`) and comparing `After` and `LastMessageId`. This check is implemented only for user tokens

Skip thread export for channel if:
- `channel.Kind == ChannelKind.GuildCategory` - categories never have threads
- `channel.Kind.IsVoice()` - voice channels never have threads
- `channel.LastMessageId` - if the channel is empty, the channel doesn't have threads (if the channel is forum post, `LastMessageId` will be `channel.Id` of the last forum post)
- (`--before`) `Before < channel.Id` - if the channel is created in the future 

Skip export of these ordinary channels if:
- `After > channel.LastMessageId` - the channel doesn't have new messages (BUT it can have new messages in threads, that's why we don't skip it in thread exports)

---

Potential problems:
- channels skipped this way won't be printed in section `Failed to export XXX channel(s):`
- Old bug - timestamp supplied using `--before` can overflow into the future - for example `--before "2010-08-30T00:00:00.000Z"` overflows to snowflake `16548918821562351616` (`2140-01-12T07:35:11.104Z`). By adding more early checks, there may be more cases, where this bug can be an issue

Tested - CLI json export using `exportGuild` with `--include-threads All` and combinations of `--before` and `--after` parameters

### DIFF
--- a/DiscordChatExporter.Cli/Commands/ExportAllCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/ExportAllCommand.cs
@@ -66,22 +66,6 @@ public class ExportAllCommand : ExportCommandBase
                     if (!IncludeVoiceChannels && channel.Kind.IsVoice())
                         continue;
 
-                    // if --after is specified, skip channels with no new messages
-                    if (
-                        After != null
-                        && channel.LastMessageId != null
-                        && After > channel.LastMessageId
-                    )
-                        continue;
-
-                    // if --before is specified, skip channels created after the specified date
-                    if (Before != null && Before < channel.Id)
-                        continue;
-
-                    // skip forums, they are exported as threads
-                    if (channel.Kind == ChannelKind.GuildForum)
-                        continue;
-
                     channels.Add(channel);
                 }
 
--- a/DiscordChatExporter.Cli/Commands/ExportAllCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/ExportAllCommand.cs
@@ -66,6 +66,22 @@ public class ExportAllCommand : ExportCommandBase
                     if (!IncludeVoiceChannels && channel.Kind.IsVoice())
                         continue;
 
+                    // if --after is specified, skip channels with no new messages
+                    if (
+                        After != null
+                        && channel.LastMessageId != null
+                        && After > channel.LastMessageId
+                    )
+                        continue;
+
+                    // if --before is specified, skip channels created after the specified date
+                    if (Before != null && Before < channel.Id)
+                        continue;
+
+                    // skip forums, they are exported as threads
+                    if (channel.Kind == ChannelKind.GuildForum)
+                        continue;
+
                     channels.Add(channel);
                 }
 
@@ -76,6 +92,8 @@ public class ExportAllCommand : ExportCommandBase
                         var thread in Discord.GetGuildThreadsAsync(
                             guild.Id,
                             ThreadInclusionMode == ThreadInclusionMode.All,
+                            Before,
+                            After,
                             cancellationToken
                         )
                     )
--- a/DiscordChatExporter.Cli/Commands/ExportGuildCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/ExportGuildCommand.cs
@@ -44,8 +44,21 @@ public class ExportGuildCommand : ExportCommandBase
             if (!IncludeVoiceChannels && channel.Kind.IsVoice())
                 continue;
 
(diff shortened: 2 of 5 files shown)


DiscordChatExporter.Cli/Commands/ExportAllCommand.cs
```diff
@@ -66,6 +66,22 @@ await foreach (var guild in Discord.GetUserGuildsAsync(cancellationToken))
...
+                    if (
+                        After != null
+                        && channel.LastMessageId != null
+                        && After > channel.LastMessageId
+                    )
+                        continue;
+
+                    // if --before is specified, skip channels created after the specified date
+                    if (Before != null && Before < channel.Id)
+                        continue;
```
Tyrrrz, line 79: We do have these checks in `ChannelExporter`, don't we?

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -277,6 +277,8 @@ public class DiscordClient
     public async IAsyncEnumerable<Channel> GetGuildThreadsAsync(
         Snowflake guildId,
         bool includeArchived = false,
+        Snowflake? Before = null,
+        Snowflake? After = null,
```
Tyrrrz, line 281: Should be lower-case

slatinsky: renamed in the last commit

DiscordChatExporter.Cli/Commands/ExportAllCommand.cs
slatinsky: You are right. I reverted checks in `ExportAllCommand` and `ExportGuildCommand` and added missing --before check to `ChannelExporter`

DiscordChatExporter.Cli/Commands/ExportGuildCommand.cs
```diff
@@ -46,6 +46,7 @@ await foreach (var channel in Discord.GetGuildChannelsAsync(GuildId, cancellatio
 
             channels.Add(channel);
         }
+        // return;
```
Tyrrrz, line 49:
```suggestion
```

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -306,14 +320,29 @@ public class DiscordClient
...
+
                     foreach (
                         var threadJson in response.Value.GetProperty("threads").EnumerateArray()
                     )
                     {
-                        yield return Channel.Parse(threadJson, channel);
+                        var thread = Channel.Parse(threadJson, channel);
+
+                        // if --after is specified, we can break early, because the threads are sorted by last message time
+                        if (after != null && after > thread.LastMessageId)
```
Tyrrrz, line 332:
```suggestion
                        if (after is not null && after > thread.LastMessageId)
```

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -338,14 +369,29 @@ public class DiscordClient
...
+
                         foreach (
                             var threadJson in response.Value.GetProperty("threads").EnumerateArray()
                         )
                         {
-                            yield return Channel.Parse(threadJson, channel);
+                            var thread = Channel.Parse(threadJson, channel);
+
+                            // if --after is specified, we can break early, because the threads are sorted by last message time
+                            if (after != null && after > thread.LastMessageId)
```
Tyrrrz, line 381:
```suggestion
                            if (after is not null && after > thread.LastMessageId)
```

Tyrrrz:
Thanks!

Tyrrrz:
> * Old bug - timestamp supplied using `--before` can overflow into the future - for example `--before "2010-08-30T00:00:00.000Z"` overflows to snowflake `16548918821562351616` (`2140-01-12T07:35:11.104Z`). By adding more early checks, there may be more cases, where this bug can be an issue

Can you please also make a PR for that in `Snowflake.FromDate(...)` method (or wherever applicable)? We can assume that a valid snowflake should be at least after Discord was launched.

PR #1138, opened 2023-10-08
Remove CDN signature before hashing asset URL

slatinsky:
Discord recently introduced signed cdn links. This change breaks `--reuse-media` for assets, because url is used to detect if the file is already downloaded.

This pull request solves this issue by stripping out `ex`, `is` and `hm` query params before hashing the url.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
@@ -7,6 +7,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Web;
 using AsyncKeyedLock;
 using DiscordChatExporter.Core.Utils;
 using DiscordChatExporter.Core.Utils.Extensions;
@@ -101,12 +102,24 @@ internal partial class ExportAssetDownloader
 
 internal partial class ExportAssetDownloader
 {
-    private static string GetUrlHash(string url) =>
-        SHA256
+    private static string GetUrlHash(string url)
+    {
+        // Strip out ex, is and hm query params to create a consistent hash
+        var uriBuilder = new UriBuilder(url);
+        var query = HttpUtility.ParseQueryString(uriBuilder.Query);
+        query.Remove("ex");
+        query.Remove("is");
+        query.Remove("hm");
+        uriBuilder.Query = query.ToString();
+        url = uriBuilder.ToString();
+        url = url.Replace(":443", "");
+
+        return SHA256
             .HashData(Encoding.UTF8.GetBytes(url))
             .ToHex()
             // 5 chars ought to be enough for anybody
             .Truncate(5);
+    }
 
     private static string GetFileNameFromUrl(string url)
     {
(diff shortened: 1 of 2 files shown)


DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -101,12 +102,24 @@ await using (var output = File.Create(filePath))
...
+        var uriBuilder = new UriBuilder(url);
+        var query = HttpUtility.ParseQueryString(uriBuilder.Query);
+        query.Remove("ex");
+        query.Remove("is");
+        query.Remove("hm");
+        uriBuilder.Query = query.ToString();
+        url = uriBuilder.ToString();
+        url = url.Replace(":443", "");
+
+        return SHA256
```
Tyrrrz, line 117: Let's do it only for Discord CDN URLs. These parameters can have some other meaning on other hosts.

slatinsky:
check for discord cdn added. Also added (simple) check if the url is really signed by checking for `ex` `is` and `hm` substrings (could be improved with regex if needed). For example avatars and emojis are not signed.

Also fixed formatting error introduced by the previous PR.

DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -101,12 +102,32 @@ await using (var output = File.Create(filePath))
...
+        {
+            var uriBuilder = new UriBuilder(url);
+            var query = HttpUtility.ParseQueryString(uriBuilder.Query);
+            query.Remove("ex");
+            query.Remove("is");
+            query.Remove("hm");
+            uriBuilder.Query = query.ToString();
+            url = uriBuilder.ToString();
+            url = url.Replace(":443", "");
+        }
```
Tyrrrz, line 123:
```suggestion
        // Strip out signature parameters from Discord CDN URLs
        if (url.Contains("cdn.discordapp.com"))
        {
            var query = HttpUtility.ParseQueryString(uriBuilder.Query);
            query.Remove("ex");
            query.Remove("is");
            query.Remove("hm");

            url = new UriBuilder(url)
            {
                Query = query.ToString()
            }.ToString();
        }
```

DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -101,12 +102,32 @@ await using (var output = File.Create(filePath))
...
+        )
+        {
+            var uriBuilder = new UriBuilder(url);
+            var query = HttpUtility.ParseQueryString(uriBuilder.Query);
+            query.Remove("ex");
+            query.Remove("is");
+            query.Remove("hm");
+            uriBuilder.Query = query.ToString();
+            url = uriBuilder.ToString();
+            url = url.Replace(":443", "");
```
Tyrrrz, line 122: Is this required? 🤔

slatinsky:
```cs
url = url.Replace(":443", "");
```
this is needed, because `Uribuilder` for some reason adds a port to the url
`https://cdn.discordapp.com/attachments/...` -> `https://cdn.discordapp.com:443/attachments/...`

The goal is to transform the url to the format as the one before Discord added signed urls.

Tyrrrz:
Thanks!

PR #1167, opened 2023-11-30
Fix GUI link in docs going to CLI page

LostXOR:
The "Using the GUI" link in the main docs page goes to the "Using the CLI" page, this PR fixes that.


Tyrrrz:
Thanks!

PR #1170, opened 2023-12-06
Add "%r" placeholder to wrap parts of the path

### LINKED ISSUE #1123: Make template tokens work properly for export paths of threads
### Version

v2.40.4

### Flavor

### Platform

Any

### Export format

_No response_

### Steps to reproduce

### Details

`\\%G\\%P – %T\\%p – %C.html`

It essentially means:

```
Server name
|
|-- Category name 1
|     |
|     |-- Channel name 1 
|     |-- Channel name 2
|     |-- Thread of channel 1
```

```
Server name
|
|-- Category name 1
|     |
|     |-- Channel name 1 
|     |
|     |-- Channel name 2
|
|-- Channel name 1
      |
      |-- Thread of channel 1
```

Possible solutions would be:

### Checklist

Paullitsch:
…be set recursively. #1123. e.g.: "\\%G\\%r%P – %T%r\\%p – %C.html"

Closes #1123

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportRequest.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportRequest.cs
@@ -153,8 +153,23 @@ public partial class ExportRequest
         Channel channel,
         Snowflake? after,
         Snowflake? before
-    ) =>
-        Regex.Replace(
+    )
+    {
+        var recursivePattern = Regex.Match(path, "%r(%.*)%r");
+        if (
+            channel.Parent?.Parent != null
+            && recursivePattern.Success
+            && recursivePattern.Groups.Count > 1
+        )
+        {
+            var groupValue = recursivePattern.Groups[1].Value;
+            path = path.Replace(
+                recursivePattern.Value,
+                FormatPath(groupValue, guild, channel.Parent, after, before)
+            );
+        }
+
+        return Regex.Replace(
             path,
             "%.",
             m =>
@@ -187,10 +202,12 @@ public partial class ExportRequest
                                 .ToString("yyyy-MM-dd", CultureInfo.InvariantCulture),
 
                         "%%" => "%",
+                        "%r" => string.Empty,
                         _ => m.Value
                     }
                 )
         );
+    }
 
     private static string GetOutputBaseFilePath(
         Guild guild,


Paullitsch:
Sorry - this lacks a little description.

The idea was to evaluate some parts of the export path-schema recursively (iterating through the parent channels).
See #1123 - here it was discussed that we can not use "%T" to put Threads into the same folder as the owner Channel. 
With this recursive call you can achieve this now because the code will update the output path as the following: 

Inital path pattern: \%G\%P – %T\%p – %C.html
updated path pattern: \%G\%r%P – %T%r\%p – %C.html

so on the first iteration will find the regex group: %P – %T
the value of the group will be forwarded again to "FormatPath" with the parent-channel as context - which than look like: 
_\%G\{Channel.Parent.Parent.Position} - {Channel.Parent.Parent.Name}\%p – %C.html_

the updated path will than be used to replace the leftover placeholder values:
_\{Guild.Name}\{Channel.Parent.Parent.Position} - {Channel.Parent.Parent.Name}\{Channel.Position} – {Channel.Name}.html_

After looking through it today again this could be more optimized but i think the conceptional idea / solution is still fine

CanePlayz:
Thanks for the thorough explanation, interesting approach. Would this mean that with only one path parameter you could end up with such a result?

```
Server name
|
|-- Category name 1
      |
      |-- Channel name 1 
      |
      |-- Channel name 2
      |
      |-- Channel name 1
            |
            |-- Thread of channel 1
```

It would after all still limit the options of how you can format the parent channel part since its structure would be inherited from the category structure, but I don't think that would be an issue as most people have these symmetrical approaches anyways.

Tyrrrz:
@CanePlayz can you comment whether this PR fully addresses #1123, as reported by you?

CanePlayz:
> This puts the threads into the same folder as the parent channel with the thread having the same naming structure as its parent channel. Rather than creating another folder inside the same directory as that of the parent channel, which is what I had imagined in the initial issue and above.

Unfortunately not. I guess it'd still be an enhancement, but:

This new method puts the thread files into the same folder as the parent channel files with the thread having the same naming structure as its parent channel. What I had imagined in the initial issue was creating another folder inside the same directory as that of the parent channel where all the threads would be stored (you can see that desired structure above).

CanePlayz:
However, a possible solution to this could be introducing new parameters that are thread-exclusive (meaning if the channel being processed is a normal one, they are just being ignored, but if the channel is a thread, they are indeed being resolved to for example the name of the parent channel or its position.

CanePlayz:
I've implemented these two parameters now. Should I open a PR for it?

Tyrrrz:
I think this is an interesting design, but I don't want to make placeholders too complicated. Adding parametrized recursion opens the door to a fully programmatic DSL inside filename templates, which could definitely be useful, but not necessarily something I want to support. I think we should offer basic templating support, but mostly defer to bulk renaming tools to let the user establish more tailored file structures.

Issue #1231, opened 2024-05-12
`--reuse-media` creates an unacceptably high chance of a 


Lymia:
### Version

v2.42.8

### Flavor

GUI (Graphical User Interface), CLI (Command-Line Interface)

### Platform

Linux

### Export format

HTML, TXT, JSON, CSV

### Steps to reproduce

Export servers where large numbers of videos, images and especially screenshots are often posted. Observe that files named `image.png` or `unknown.png` are extremely common due to copy-paste of images directly into Discord, often leaving no file name available.

Do math to confirm the birthday problem is at play and will be causing you problems.

### Details

The algorithm for making file names for downloaded media only stores 20 bytes of the hash of the URL. This is insufficient.

Despite there being 1048576 possible hashes, it only takes 1200 images for there to be a 50% chance of a collision and 500 images before there is a 10% chance of a collision due to the [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem). The original reasoning in the issue where this code was added (#395) is incorrect due to the commonness of `unknown.png`, `image.png`, and other filenames.

A list of all filenames I've seen that *could* reach that 10% chance of a collision in just my own exports of servers I'm in:
```
    508 hqdefault.jpg
    541 image0.png
    681 sddefault.jpg
   1269 image0.jpg
   1696 tenor.png
   5798 maxresdefault.jpg
   6637 image.png
  11314 unknown.png
```
As can clearly be seen, file names are not evenly distributed. Notably, `maxresdefault`, `sddefault` and `hqdefault` are not being posted intentionally by users at all, and are instead the result of youtube thumbnails. Similarly, `tenor` happens because the tenor gif button is pressed in Discord's own UI.

There is a near 100% chance that there are collisions here between `unknown.png`, `image.png`, and `maxresdefault.jpg`. You can check with this calculator: https://www.bdayprob.com/. Solve `P(D,N)` with `D = 1048576` and `N = 6637`.

10 characters provides a better margin of safety: 15000 images with the same name before there is even a 0.01% chance, which would have successfully handled my use case.

10 characters of base-32 rather than base-16 would provide 600000 images before there is the same chance of collision. 8 characters of base-32 would provide the same margin as the 10 characters of base-16.

### Checklist

- [X] I have looked through existing issues to make sure that this bug has not been reported before
- [X] I have provided a descriptive title for this issue
- [X] I have made sure that this bug is reproducible on the latest version of the application
[...]

PR #1232, opened 2024-05-13
Improve filenames for downloaded assets.

Lymia:
This changes the filenames output by DiscordChatExporter to be more meaningful (in the case of emojis and avatars) and more resilient against collisions.

The hash is 12 characters of base32 now, rather than 5 characters of hexadecimal. This allows for nearly 5 million downloads with the same name before there is even a 0.001% of a single collision. This should *actually* be enough, even for problematic filenames like those associated with Youtube thumbnails.

Emojis and Discord attachments are instead *guaranteed* to not collide as they instead contain the unique snowflake of the attachment or emoji.

The 19 digit id is significantly longer than the old 5 digit ids - however, reencoding it in base32 would only save 6 characters, so better to use the more recgonizable numeric form IMO.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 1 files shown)


Lymia:
Any update on this? I personally do use this patch to backup a few servers that trigger this bug.

Tyrrrz:
Can't really change the naming algorithm as it will break "skip downloaded assets" for all existing exports.

Lymia:
That is, unfortunately, the point. "Skip Downloaded Assets" as it currently works has a severe risk of incorrectly skipping non-duplicated assets, and there's no good way to fix that without invalidating past downloads. Do you think it'd work to put this behind a command line flag, instead of changing the default behavior?

Tyrrrz:
I think your concern is statistically valid, but there yet haven't been any bug reports in regard to asset filename collision. If/when that happens, then it makes sense to introduce the breaking change to fix such an issue. I want to make sure we're fixing real problems when we add new ones (by breaking compatibility) 🙂 

Maybe down the line there will be other related breaking changes, then it will make sense to group them together. Right now I don't see it happening any time soon.

Tyrrrz:
> would there be an error? or would the file just silently be skipped? in a bigger discord that would be very easy to miss if it was just silently skipped

There wouldn't be an error now, but that can be changed.

Twi-Hard:
I have a single channel with 8774 conflicts.
This command counts how many times each filename occurred:
```
❯ cat 98609319519453184.json | jq -r '.messages[].attachments[].url' | sort | uniq -c | sort -nr > discord-98609319519453184-dupes.txt
```
This command removes the lines with only 1 instance of a filename then sums up all of the file counts for files with conflicts:
```
❯ cat discord-98609319519453184-dupes.txt | awk '{print $1}' | grep -vE '^1$' | paste -sd+ - | bc
17453
```
Then count the number of filenames that occur more than once:
```
❯ cat discord-98609319519453184-dupes.txt | awk '{print $1}' | grep -vE '^1$' | wc -l
8679
```
17453-8679 = 8774

I manually checked some of the messages that had duplicate filenames in discord and they were indeed unique images in discord but only image was saved locally.
I've checked a different channel too and the same thing happened.
I can't remember if it's the default or not, but I store each channels images in its own folder
```
❯ ls -1U | head -10
670866322619498511-media
670866322619498511.json
704107851421057114-media
704107851421057114.json
704226060178292846-media
704226060178292846.json
862221868621758502-media
862221868621758502.json
862732083050971146-media
862732083050971146.json
```
It would be really great if you could implement an option to make each name unique. People that want to stick with the old method don't need to do anything if it's done this way.

Tyrrrz:
@Twi-Hard please add this info to the linked issue (#1231). If you can, please also provide a few different URLs that resolve to the same hash/filename with the current algorithm.

Tyrrrz:
Just as an update, this fix will be included in `3.x` as part of #1271. I just can't promise when that will be.

WanHaes:
How can I use this patch now?

PR #1233, opened 2024-05-13
Use a 3-way theme switcher instead of a 2-way switcher

Tyrrrz:


96-LB:
I completely misread the changes, and for a second I thought you added a new DeepOrange-colored HTML export 🍊

PR #1254, opened 2024-06-27
Remove unnecessary prefixes from CLI docs

LostXOR:
The CLI docs have `$` and `./` prefixes before each command. This is confusing, as the `$` is not a part of the command at all and the `./` is Unix-specific.

This PR removes the prefixes from the commands, and adds a note at the beginning explaining that the `./` prefix should be added for MacOS/Linux users.

### DIFF
--- a/.docs/Using-the-CLI.md
+++ b/.docs/Using-the-CLI.md
@@ -18,14 +18,17 @@ You can also drag and drop the folder on **every platform**.
 
 ## Step 3
 
-Now we're ready to run the commands. The examples on this page follow the Windows file path format, change the file
-paths according to your system.
+Now we're ready to run the commands.
 
 Type the following command in your terminal of choice, then press ENTER to run it. This will list all available subcommands and options.
 
 ```console
-$ ./DiscordChatExporter.Cli
+DiscordChatExporter.Cli
 ```
+
+> **Note**:
+> The examples on this page use the **Windows** file path format and command syntax. **MacOS** and **Linux** users will need to prefix commands with `./` (the previous command becomes `./DiscordChatExporter.Cli`) and use Unix file paths.
+
 
 > **Docker** users, please refer to the [Docker usage instructions](Docker.md).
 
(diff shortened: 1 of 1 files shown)


Tyrrrz:
`./` is also required on Windows with PowerShell if I remember correctly. Can you check?

LostXOR:
I'm not sure, but the docs specify to use Command Prompt, and `./` doesn't work with it.

> After extracting the .zip, open Command Prompt, aka cmd (Terminal on macOS and Linux).

CharithBhat:
@Tyrrrz @LostXOR 
Tried it on Windows Powershell. `./` is required. 
But with Command Prompt. I have to remove the `./` for it to work.

Tyrrrz:
So it seems that all shells except Command Prompt require the `./`, while Command Prompt doesn't work with `./`. It's annoying. Maybe it's better to keep the `./` in docs and just add a note for Command Prompt users?

LostXOR:
I guess that's probably best, I've updated the docs accordingly.

.docs/Using-the-CLI.md
```diff
@@ -18,15 +18,18 @@ You can also drag and drop the folder on **every platform**.
...
 
 Type the following command in your terminal of choice, then press ENTER to run it. This will list all available subcommands and options.
 
 ```console
-$ ./DiscordChatExporter.Cli
+./DiscordChatExporter.Cli
 ```
 
+> **Note**:
+> Command Prompt users must omit the `./` from the start of the command. The examples on this page use **Windows** file paths. **MacOS** and **Linux** users must use Unix file paths accordingly.
```
Tyrrrz, line 30: Simplified tone.

```suggestion
> On Windows, if you're using the default Command Prompt (`cmd`), omit the leading `./` at the start of the command.
```

.docs/Using-the-CLI.md
```diff
@@ -2,7 +2,7 @@
 
 ## Step 1
 
-After extracting the `.zip`, open Command Prompt, aka `cmd` (`Terminal` on **macOS** and **Linux**).
+After extracting the `.zip`, open your terminal (Command Prompt or Powershell on **Windows**, and Terminal on **macOS** and **Linux**).
```
Tyrrrz, line 5: Removed the extra details because they introduce confusion between terminal/shell concepts. On Windows, users should use the Terminal app ideally as well, but that is not too important.

```suggestion
After extracting the `.zip` archive, open your preferred terminal.
```

Tyrrrz (approved):

PR #1288, opened 2024-09-13
Fix issue with threads named 'deleted-channel'

### LINKED ISSUE #1261: Thread names get exported as `#deleted-channel` in mentions
### Version

2.43.3

### Flavor

CLI (Command-Line Interface)

### Platform

Linux (Debian)

### Export format

HTML

### Steps to reproduce

- Make a mention to a thread in a channel
- Export that channel
- Observe the mention being converted to `#deleted-channel`

### Details

This piece of code is relevant;

https://github.com/Tyrrrz/DiscordChatExporter/blob/281e0f608ad0784bce1dc2d2cc7b76e419905ebd/DiscordChatExporter.Core/Exporting/HtmlMarkdownVisitor.cs#L284

It only looks for a name of the channel, not of a thread.

Even discord seems to struggle, not fetching a corresponding thread, only saying `#unknown`

### Checklist

maxgsf:
Closes #1261

I tried to solve the problem with the thread labeled 'deleted-channel' by using populate threads before assigning the channel variable in the mention.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportContext.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportContext.cs
@@ -48,21 +48,26 @@ internal class ExportContext(DiscordClient discord, ExportRequest request)
         }
     }
 
-    public async ValueTask PopulateThreadsAsync(CancellationToken cancellationToken = default)
+    public async ValueTask PopulateThreadsAsync(
+        Snowflake? id,
+        CancellationToken cancellationToken = default
+    )
     {
-
-        await foreach (
-            var channel in Discord.GetGuildThreadsAsync(
-                Request.Guild.Id,
-                true,
-                null,
-                null,
-                cancellationToken
-            )
-        )
+        if (id == null)
         {
-            _channelsById[channel.Id] = channel;
+            return;
         }
+
+        Snowflake sid = (Snowflake)id;
+
+        if (_channelsById.ContainsKey(sid))
+        {
+            return;
+        }
+
+        var channel = await Discord.GetChannelAsync(sid, cancellationToken);
+
+        _channelsById[sid] = channel;
     }
 
     // Because members cannot be pulled in bulk, we need to populate them on demand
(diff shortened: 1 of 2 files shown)


Tyrrrz:
Thanks. This works, but pulling threads for an entire guild is notoriously slow.

maxgsf:
Thanks for the feedback! I've updated my code to only fetch the necessary channel. Please check again.

DiscordChatExporter.Core/Exporting/ExportContext.cs
```diff
@@ -48,6 +48,28 @@ var channel in Discord.GetGuildChannelsAsync(Request.Guild.Id, cancellationToken
...
+
+        if (_channelsById.ContainsKey(sid))
+        {
+            return;
+        }
+
+        var channel = await Discord.GetChannelAsync(sid, cancellationToken);
+
+        _channelsById[sid] = channel;
+    }
```
Tyrrrz, line 71: Check for `null` at callsite instead

```suggestion
    public async ValueTask PopulateThreadAsync(
        Snowflake id,
        CancellationToken cancellationToken = default
    )
    {
        if (_channelsById.ContainsKey(id))
            return;

        var channel = await Discord.GetChannelAsync(id, cancellationToken);

        _channelsById[id] = channel;
    }
```

DiscordChatExporter.Core/Exporting/HtmlMarkdownVisitor.cs
```diff
@@ -279,6 +279,8 @@ protected override async ValueTask VisitMentionAsync(
         }
         else if (mention.Kind == MentionKind.Channel)
         {
+            await context.PopulateThreadsAsync(mention.TargetId, cancellationToken);
```
Tyrrrz, line 282:
```suggestion
        if (mention.TargetId is not null)
            await context.PopulateThreadsAsync(mention.TargetId, cancellationToken);
```

Tyrrrz:
Looks like this PR has been abandoned

PR #1311, opened 2024-10-26
Include inline emoji in JSON export

### LINKED ISSUE #1026: JSON option does not seem to export in-line emoji which are from other servers
### Version

2.39.1

### Flavor

### Export format

JSON

### Details

When exporting a channel in HTML, all emoji are preserved exactly as they are. When exporting the same channel in JSON, built-in Discord emoji and emoji specific to the server are preserved, as are all emoji in reactions, but emoji which are both embedded in the text of a message _and_ originate from a different server do not seem to be preserved, and simply display in the message as :emoji:.
Due to the way the media files are named I have no way to actually know if they're being downloaded or not, and I also don't really know how the exports are supposed to associate the regular in-line :emoji: with their images since they don't have any explicit imageUrl like the reactions do, but every other emoji is showing up fine and these ones aren't.

Edit: Actually upon further testing, _sometimes_ emoji from other servers display correctly in-line, no idea why, and I've come across at least one instance of an emoji from a different server being rendered in chat, but as a completely different emoji.
I've also seen one instance of an emoji from the exported server failing to render, but just a few messages earlier someone had posted exactly the same emoji with a different name from a different server (which rendered successfully.) I thought it might have something to do with the "reuse assets" toggle, but I disabled it and tried the export again and saw no change.

### Steps to reproduce

- Download section of Discord channel that has emoji which originate from a different server posted in messages as part of the message text, as well as in reactions to messages.
- Open export in a frontend.
- External emoji are shown as plaintext :emoji: while all other emoji show up fine, including the external emoji in reactions.

Tyrrrz:
Closes #1026

- Adds `messages[].inlineEmojis[]` that contains an array of emoji objects, describing emoji used inside `{message}.content`.
- Adds `messages[].embeds[].inlineEmojis[]` that contains an array of emoji objects, describing emoji used inside `{embed}.description`.
- Both arrays are deduplicated and may be empty but never null.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
+++ b/DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using DiscordChatExporter.Core.Discord.Data;
 using DiscordChatExporter.Core.Discord.Data.Embeds;
+using DiscordChatExporter.Core.Markdown.Parsing;
 using DiscordChatExporter.Core.Utils.Extensions;
 using JsonExtensions.Writing;
 
@@ -25,7 +26,7 @@ internal class JsonMessageWriter(Stream stream, ExportContext context)
                 Indented = true,
                 // Validation errors may mask actual failures
                 // https://github.com/Tyrrrz/DiscordChatExporter/issues/413
-                SkipValidation = true
+                SkipValidation = true,
             }
         );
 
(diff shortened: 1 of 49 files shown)


DiscordChatExporter.Core/Exporting/JsonMessageWriter.cs
```diff
@@ -512,11 +513,23 @@ await Context.ResolveAssetUrlAsync(
...
 
             _writer.WritePropertyName("user");
-            await WriteUserAsync(message.Interaction.User, cancellationToken);
+            await WriteUserAsync(message.Interaction.User, true, cancellationToken);
 
             _writer.WriteEndObject();
         }
 
+        // Inline emoji
+        _writer.WriteStartArray("inlineEmojis");
```
Tyrrrz, line 522: Ideas for a better name?

slatinsky:
Hello,

I have tested this pull requests and I have found some small issues:

1. `messages[].embeds[].description` sent by automod or by bots in a embed may contain custom emojis. Those emojis don't show up in `messages[].inlineEmojis[]`
2. the same emoji sent multiple times in the same `messages[].content` is not deduplicated. 

Other than that, the pull request looks good to me. Including the name `inlineEmojis`

Tyrrrz:
> Hello,
> 
> I have tested this pull requests and I have found some small issues:
> 
> 1. `messages[].embeds[].description` sent by automod or by bots in a embed may contain custom emojis. Those emojis don't show up in `messages[].inlineEmojis[]`
> 2. the same emoji sent multiple times in the same `messages[].content` is not deduplicated.
> 
> Other than that, the pull request looks good to me. Including the name `inlineEmojis`

Good points.

As for 1., do you think those emoji should be inside `messages[].inlineEmojis[]` or `messages[].embeds[].inlineEmojis[]`? The former is obviously easier, but I wonder if there's any downside.

PR #1343, opened 2025-02-20
Channel threads

### LINKED ISSUE #1274: --include-threads not available for `export` command
### Version

v2.43.3

### Flavor

CLI (Command-Line Interface)

### Platform

MacOS

### Export format

_No response_

### Steps to reproduce

```
./DiscordChatExporter.Cli export -t "$DISCORD_TOKEN" -c 621778831602221072 --after "2024-08-13 00:00" --before "2024-08-15 00:00" -f PlainText -o ~/Desktop/%G-%c-%a-%b-PlainText --include-threads
Unrecognized option(s):
--include-threads

DiscordChatExporter.Cli v2.43.3

USAGE
  DiscordChatExporter.Cli export --channel <values...> --token <value> [options]
```

### Details

This means there is no way to export only a particular channel in the guild and at the same time include threads. As a result if I want to capture threads I have to run a much bigger export job with potentially 10s or 100s of irrelevant channels, taking much longer and risking getting my token throttled or blacklisted.

### Checklist

audinowho:
Closes #1274

Kinda.

I wanted to be able to pull all threads under a channel, when running `export` with a channel or forum.
Output still alerts when pulling a forum channel, even though the threads were successfully grabbed.
Not fully tested, but has served me well in my own runs.
Hope it's helpful in some fashion.  Open to suggestions.


wxiaoyun:
Thanks a lot!

I tried it on my machine with Docker and it works!

DiscordChatExporter.Cli/Commands/ExportChannelsCommand.cs
```diff
@@ -19,9 +26,86 @@ public class ExportChannelsCommand : ExportCommandBase
...
     public required IReadOnlyList<Snowflake> ChannelIds { get; init; }
 
+    [CommandOption(
+        "include-threads",
+        Description = "Which types of threads should be included.",
+        Converter = typeof(ThreadInclusionModeBindingConverter)
+    )]
+    public ThreadInclusionMode ThreadInclusionMode { get; init; } = ThreadInclusionMode.None;
+
     public override async ValueTask ExecuteAsync(IConsole console)
```
Tyrrrz, line 36: Why not perform this in the `ExportCommandBase` instead? Currently, that's where we unwrap categories already.

audinowho: Wasnt sure if it made sense to be in all children but looks like they all handle it in some way.  moved it.

Tyrrrz: Oh I meant also moving the unwrapping logic from `ExportChannelsCommand.ExportAsync(...)` to `ExportCommandBase.ExportAsync(...)`

Tyrrrz:
Note that the code responsible for pulling threads in `DiscordClient` has recently been changed to handle pagination correctly, please pull in the master branch and see if your new method `GetChannelThreadsAsync(...)` adapts the changes correctly.

Tyrrrz:
@audinowho are you still working on this? Do you need help?

audinowho:
I've verified against the latest commit and the new pagination code.
I'm somewhat unsure on how to move the unwrapping logic from `ExportChannelsCommand.ExportAsync(...)` to `ExportCommandBase.ExportAsync(...)` though.
- `ExportChannelsCommand` calls `GetChannelThreadsAsync(...)`
- `ExportGuildCommand` calls `GetGuildThreadsAsync(...)`
- `ExportAllCommand` calls `GetGuildThreadsAsync(...)` multiple times.
- `ExportDirectMessagesCommand` does not call anything thread related.
I could move the thread logic in ExportChannelsCommand to the beginning of `ExportCommandBase.ExportAsync(...)` and remove it from all other classes.  However, this will require changing `ExportAsync` to pass a cancellation token.  Is that ok?

DiscordChatExporter.Cli/Commands/ExportChannelsCommand.cs
audinowho: Added now.

audinowho:
nvm, figured it out.  Current code tested to work for all child classes now.

PR #1349, opened 2025-03-03
Don't consider it an error if there is nothing to export

### LINKED ISSUE #1327: No new messages since `--after MESSAGE_ID` should be considered a warning and exit status should be zero
### Version

v2.44

### Flavor

CLI (Command-Line Interface)

### Platform

Docker / Linux

### Export format

JSON

### Steps to reproduce

1. `export --after MESSAGE_ID` for a channel that has no new messages since given ID
2. exit code is non-zero even though this isn't strictly an error condition

### Details

I am using the `--after` flag to do incremental scraping, passing the last message ID seen in previous runs to get only new messages since then.

The problem is that no new messages results in non-zero exit status, indistinguishable from other more serious errors without resorting to brittle parsing of console output.

A saner behavior would be to treat this as a warning instead of an error, and exit with zero status instead, so that external scripting of the CLI can be simple and reliable.

### Checklist

ldmosquera:
With `--before` or `--after`, if there are no new messages in the given period then an empty export file will be produced with a successful exit, rather than an error exit.

Closes #1327


ldmosquera:
Not tested; keeping as draft while I work out how to test this locally.

ldmosquera:
TODO:
- [x] successful exit code if no messages
- [ ] empty file is not written out as I assumed it would be; no file is created. Need to make it so file is created as discussed
- [ ] console output is currently iffy; need to figure out how to cleanly output errors

tags
```diff
@@ -0,0 +1,871 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
Tyrrrz, line 1: Probably don't need to commit this

ldmosquera:
>  console output is currently iffy due to need for async; need to figure out how to cleanly output to STDERR

The issue is that `ExportChannelAsync` is meant to be headless and used indistinctly from CLI and GUI, and a console is not available in the latter. Plus, right now the execution for a channel is either flawless + silent, or aborted with an exception whose message is the error message to be shown. There is no support for it to succeed but also return warning messages to be shown.

@Tyrrrz any guidance on how to proceed?
- the cheap solution would be to optionally take an `IConsole` object when called from CLI and use that only from CLI (silently ignoring the situation in GUI)
- the correct solution would be to add generic support for somehow returning "warning messages", and then show them at the end of the run similarly to error messages from catched exceptions
- the non-solution would be to just be silent about this condition either way. IMHO this edge case doesn't warrant too much refactoring

Tyrrrz:
@ldmosquera I think the best way to go about it is to keep the exception throw, but introduce a more specific type `DiscordChatExporterException -> ChannelEmptyException`. Then handle that exception on CLI/GUI side to produce an error message without failing.

ldmosquera:
Good call; note that for consistency, I'm also handling an attempt to export an empty channel as a non-error.

I made the exception one conceptual level above "channel was empty" to "no messages exported", because otherwise we would be throwing it on no __new__ messages, which isn't strictly the same thing. This way it can be used for both an empty channel and for "no new messages in given period". These are now handled the same for consistency.

I only tested the CLI output, but the GUI change is very minimal.

Updated TODOs:
[x] successful exit code if no new messages
[ ] empty file is not written out as I assumed it would be; no file is created. Need to make it so file is created as discussed
[x] console output is currently iffy due to need for async; need to figure out how to cleanly output to STDERR
[ ] color is not applied in CLI STDERR output

DiscordChatExporter.Gui/ViewModels/Components/DashboardViewModel.cs
```diff
@@ -283,6 +283,12 @@ await Parallel.ForEachAsync(
 
                         Interlocked.Increment(ref successfulExportCount);
                     }
+                    catch (ChannelNotExportedException ex) {
+                        _snackbarManager.Notify(ex.Message.TrimEnd('.'));
+
+                        // FIXME: not exactly successful, but not a failure either. Not ideal to duplicate the line
+                        Interlocked.Increment(ref successfulExportCount);
```
ldmosquera, line 290: This could also be handled in line 284, thus not needing to duplicate the line, but then error handling would be spread around inside of at the obvious place. Neither is 100% clean.

DiscordChatExporter.Core/Exporting/MessageExporter.cs
```diff
@@ -30,6 +30,11 @@ private async ValueTask ResetWriterAsync(CancellationToken cancellationToken = d
         }
     }
 
+    // Unless explicitly called, file will be created before writing first message
+    public async ValueTask EnsureFileIsCreated(CancellationToken cancellationToken = default) {
+        await GetWriterAsync(cancellationToken);
+    }
```
ldmosquera, line 36: Might be more explicit than strictly needed, but it documents intended behavior.

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -252,26 +256,33 @@ await console.Output.WriteLineAsync(
...
-            using (console.WithForegroundColor(ConsoleColor.Red))
+            var tuples = new List<(ConcurrentDictionary<Channel, string>, ConsoleColor, string)>
             {
-                await console.Error.WriteLineAsync(
-                    $"Failed to export {errorsByChannel.Count} the following channel(s):"
-                );
-            }
+                (errorsByChannel,   ConsoleColor.Red,    "Failed to export the following channels:"),
+                (warningsByChannel, ConsoleColor.Yellow, "No messages exported for the following channels:")
+            };
```
Tyrrrz, line 266: I'd rather split these into two similar code blocks instead (one for errors, one for warnings), for readability

DiscordChatExporter.Core/Exporting/ChannelExporter.cs
```diff
@@ -27,45 +27,35 @@ public async ValueTask ExportChannelAsync(
...
-                    + $"of guild '{request.Guild.Name}' "
-                    + $"does not contain any messages."
-            );
-        }
-
-        // Check if the 'after' boundary is valid
-        if (request.After is not null && !request.Channel.MayHaveMessagesAfter(request.After.Value))
-        {
-            throw new DiscordChatExporterException(
+            throw new ChannelNotExportedException(
```
Tyrrrz, line 41: Still sounds like `ChannelEmptyException` is an okay name, given that we're not throwing any other exception to discern between "fully empty" and "circumstantially empty"

DiscordChatExporter.Core/Exporting/MessageExporter.cs
```diff
@@ -30,6 +30,11 @@ private async ValueTask ResetWriterAsync(CancellationToken cancellationToken = d
         }
     }
 
+    // Unless explicitly called, file will be created before writing first message
+    public async ValueTask EnsureFileIsCreated(CancellationToken cancellationToken = default) {
```
Tyrrrz, line 34: It might be better to call this implicitly during disposal instead.

DiscordChatExporter.Core/Exporting/MessageExporter.cs
```diff
@@ -30,6 +30,11 @@ private async ValueTask ResetWriterAsync(CancellationToken cancellationToken = d
...
+        await GetWriterAsync(cancellationToken);
```
Tyrrrz, line 35: This would only result in a 0-byte file, right? I think the original intention was to create a fully structurally sound export, just with no messages.

Tyrrrz:
As for tests, I think you can add a case here: https://github.com/Tyrrrz/DiscordChatExporter/blob/master/DiscordChatExporter.Cli.Tests/Specs/DateRangeSpecs.cs

Where you run an export with ranges long in the past, and ensure that the export file exists but has no messages. Unfortunately, you can't run the tests locally due to their reliance on a bot token.

DiscordChatExporter.Core/Exporting/MessageExporter.cs
ldmosquera: No, `GetWriterAsync` causes `WritePreambleAsync` to be run, which means you get the expected boilerplate as you said. Test coming up which should clear it up.

Tyrrrz: Ah, cool

DiscordChatExporter.Core/Exporting/MessageExporter.cs
ldmosquera: Done in https://github.com/Tyrrrz/DiscordChatExporter/pull/1349/commits/27a7bdd8702e41ad95566a3b6797c3a52f74924a

ldmosquera:
Test added in https://github.com/Tyrrrz/DiscordChatExporter/pull/1349/commits/77c125f530f98df398211edfbd4153763e9fcff4

I haven't run it of course but it's very similar to others with only straightforward changes, and it all compiles at least.

DiscordChatExporter.Core/Exporting/MessageExporter.cs
```diff
@@ -70,7 +70,15 @@ public async ValueTask ExportMessageAsync(
         MessagesExported++;
     }
 
-    public async ValueTask DisposeAsync() => await ResetWriterAsync();
+    protected async ValueTask EnsureFileIsCreated(CancellationToken cancellationToken = default) {
+        // causes the file to be created whether there were messages written or not
+        await GetWriterAsync(cancellationToken);
+    }
```
Tyrrrz, line 76: It's better to inline this method (call `GetWriterAsync` directly) into `DisposeAsync`

DiscordChatExporter.Gui/ViewModels/Components/DashboardViewModel.cs
Tyrrrz: Can you explain what duplication you're referring to?

Tyrrrz (commented):
Nice. Let me know when it's out of draft.
There's also some formatting issues you can fix by running `dotnet build` in the repository root.

DiscordChatExporter.Core/Exporting/MessageExporter.cs
ldmosquera: The intention is more clear this way, but your call; done in https://github.com/Tyrrrz/DiscordChatExporter/pull/1349/commits/daff360ce4cf0cc7c6b9555559d0fe3da17cefe3

DiscordChatExporter.Gui/ViewModels/Components/DashboardViewModel.cs
ldmosquera: If you click into the diff link, you'll see the original line at 284 which I'm duplicating at 289. Slightly iffy but I can't find a cleaner way to express this logic.

ldmosquera:
PR is ready to go on my end; happy to resolve any other outstanding points though.

Tyrrrz:
There still seems to be one formatting error

ldmosquera:
Oops, that wasn't picked up by `dotnet format`. Fixed.

PR #1355, opened 2025-03-17
Revert to `Uri.EscapeDataString` when encoding asset URIs

### LINKED ISSUE #1353: Asset viewing in Firefox is broken
### Version

2.44.1

### Flavor

### Platform

Windows 11

### Export format

HTML

### Steps to reproduce

Export settings:
- **Download assets**: Enabled

Browser: Firefox

### Details

Expected behavior: Assets load when viewing the HTML export in Firefox
Actual behavior: Assets fail to load due to the URL-encoded file path
Caused by: #1351

### Checklist

96-LB:
Closes #1353 

After fixing #1351, a commit was made to change the encoding function from `Uri.EscapeDataString` to `WebUtility.UrlEncode`. This introduced a new bug because the latter function encodes spaces as `+`, which is a valid filename character and not interpreted correctly by the asset links. This PR rolls back the change to restore the intended behavior. I've tested that filenames which include both ` ` and `%` are properly handled.

### DIFF
--- a/DiscordChatExporter.Core/Utils/Url.cs
+++ b/DiscordChatExporter.Core/Utils/Url.cs
@@ -1,4 +1,4 @@
-﻿using System.Net;
+﻿using System;
 using System.Text;
 
 namespace DiscordChatExporter.Core.Utils;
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Thanks!

PR #1364, opened 2025-04-17
Don't ignore next pages of archived threads

### LINKED ISSUE #1251: Not all threads are fetched from a server when it has a lot of them
### Version

2.43.3

### Flavor

CLI (Command-Line Interface)

### Platform

win11

### Export format

HTML, TXT

### Steps to reproduce

When exporting threads via the --include-threads command or by exporting a list of threads using the channel command, if a server has a very large amount of threads, not all threads will download or be listed,
For example i have a server with multiple 300+ thread channels. CLI will only collect a third of it roughly.

i believe this may possibly be a discord issue aswell as when viewing those channels in gallary view via the discord client the full list of channels wont load unless you switch to list view, then they will all load. perhaps related? (to reproduce the discord client issue you must restart your client to clear cache, once all channels are loaded they will be visible in both gallary and list view but exclusively load only in list view to its entirety)

### Details

Expecting to export full list of roughly 1000 threads,
actually results are less than expected value but consistant across multiple attempts. if it only exports 100 of the 300 threads it will do that the second attempt aswell.

Attempted with both
DiscordChatExporter.Cli.exe channels -g 123456789 --include-threads all
and
exportguild -g123456789 --include-threads all

### Checklist

masmc05:
Closes #1251 
Old:
image
New:
image

Fixes fetching the threads by fetching the next pages of threads. Unlike active threads, archived aren't returned all at once.

### DIFF
--- a/DiscordChatExporter.Core/Discord/DiscordClient.cs
+++ b/DiscordChatExporter.Core/Discord/DiscordClient.cs
@@ -438,39 +438,59 @@ public class DiscordClient(string token)
                 foreach (var channel in channels)
                 {
                     // Public archived threads
-                    {
-                        // Can be null on certain channels
-                        var response = await TryGetJsonResponseAsync(
-                            $"channels/{channel.Id}/threads/archived/public",
-                            cancellationToken
-                        );
-
-                        if (response is null)
-                            continue;
-
-                        foreach (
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Thanks!

PR #1461, opened 2026-01-03
Add flag to enable collisionless media downloads

### LINKED ISSUE #1231: `--reuse-media` creates an unacceptably high chance of a

benrucker:
Closes https://github.com/Tyrrrz/DiscordChatExporter/issues/1231

> [!NOTE]
> My understanding of the original issue is that when downloading media with the `--media` flag (with or without `--reuse-media`) there is a significant chance that multiple pieces of media could attempt to use the same filename + hash, causing all but one media download to be silently lost. If this is wrong, let me know!

## Change proposed 

This PR introduces a new system for saving media downloads gated behind an opt-in flag `--new-media-paths` that eliminates the chance for media filenames to collide by using existing unique IDs from Discord.

## Rationale

This implementation takes inspiration from this PR https://github.com/Tyrrrz/DiscordChatExporter/pull/1232 and its comment https://github.com/Tyrrrz/DiscordChatExporter/issues/1231#issuecomment-2106399351:

> An alternative suggestion to lengthening the hash would be to just put the attachment ID in the filename. I feel like at a certain point, a bunch of random base-32 characters doesn't really provide much utility over 19 digits (which would be guaranteed to never collide).

This seemed reasonable to me, so I decided to look in to implementing a version of it myself for personal use.

The interesting thing I noticed is that out of the 5 Discord media types that DCE downloads via `ExportAssetDownloader` (emojis, stickers, avatars, attachments, server icons), **only three of them use snowflakes** while the other two use hashes. The two that use hashes, avatars and server icons, are uniquely identified by a combination of the member/server ID and the file hash:

| Asset | URL | Unique identifier |
| :--- | :--- | :--- |
| Custom Emoji | `emojis/{emoji_id}.png` | `{emoji_id}` |
| Sticker | `stickers/{sticker_id}.png` | `{sticker_id}` |
| Attachment | `attachments/{channel_id}/{attachment_id}/{attachment}.png` | `{attachment_id}` |
| Guild Icon | `icons/{guild_id}/{guild_icon}.png` | `{guild_id}`+`{guild_icon}` |
| Guild Member Avatar | `guilds/{guild_id}/users/{user_id}/avatars/{member_avatar}.png` | `{user_id}`+`{member_avatar}` |

So strictly speaking, if we wanted to save something using just its unique ID (snowflake or composite ID+hash), we'd have to introspect the URL, determine the type of asset we're downloading, then branch on that to properly extract the right components from the URL.

But we could take this one step further: why extract parts from the URL to construct a unique filename when the URL is already a unique identifier itself? Using the URL itself has the added benefit of naturally separating the media downloads by their type:
emojis go in the `emojis/` folder, stickers go in the `stickers/` folder, etc.

## Single-file examples

When `ExportAssetDownloader` receives a media URL, it checks if the URL is from the Discord CDN. If so, it parses the path from the rest of the URL and uses that as the path to the file that will be saved to disk. 

If the `size` query param is present, that is added as the last folder in the file path (a la Adobe Illustrator's technique for organizing exports of assets across multiple size). This prevents a possible issue where `ExportAssetDownloader` receives two requests for the same asset at different sizes causing one size to be overwritten.

### Discord CDN file
For example, if we get this URL passed: 
```
https://cdn.discordapp.com/emojis/754441880066064584.webp?size=128
```

Then we will save that media to this path on disk
```
{media_path}/emojis/128px/754441880066064584.webp
```

### External media file

If the URL is not from the discord CDN, then it's sequestered into an `external/` subfolder and then uses the path in the URL as the filepath beyond that. For example, an embed might have a URL like the following:
```
https://images-ext-1.discordapp.net/external/kJEb3-XApHsU0ov41y9Nk8BuseDcmeUSHQEtN0swR2M/https/i.ytimg.com/vi/6RghYTEwbpM/maxresdefault.jpg
```

Which would get saved to this location:
```
{media_path}/external/images-ext-1.discordapp.net/external/kJEb3-XApHsU0ov41y9Nk8BuseDcmeUSHQEtN0swR2M/https/i.ytimg.com/vi/6RghYTEwbpM/maxresdefault.jpg
```

## Wholistic directory structure

In effect, running a command that looks something like this:
```sh
./DiscordChatExporter.Cli export
  -t "wHid.spr.bNa" 
  --media 
  --reuse-media 
  --new-media-paths 
  -o "output/data/"
  --media-dir "output/media/"
```

Would result in an output structure along these lines:
```
output/
  data/
    guild name - category - channel1 [id].json
    guild name - category - channel2 [id].json
    guild name - category - channel3 [id].json
    ...
  media/
    attachments/
      173123.../ (server_id)
        145699.../ (attachment_id)
          IMG_1550.png
        ...
      ...
    avatars/
      17294.../ (user id)
        cadcc293.png (avatar hash)
      ...
    emojis/
      1234219.png (emoji id)
      ...
    external/
      cdn.jsdelivr.net/
      images-ext-1.discordapp.net/
      www.youtube.com/
      ...
    guilds/
      170123.../ (guild_id)
        users/
          17290.../ (user id)
            avatars/
              512px/
                50e477.png (avatar hash)
              ...
          ...
      ...
    icons/
      170123.../ (guild_id)
        512px/
          abde23.gif (file hash)
        ...
```

## Comparison to the old behavior

The old behavior puts all media files flat into the same folder. Some benefits of this is that it reduces the number of folders on your disk, reduces the number of characters needed to point to each file, and in a way feels clean when you see everything packaged together. A commonly cited drawback of this is that filename collisions are highly likely.

The new behavior's nesting improves accessibility of casually inspecting the results of a backup via the native file browser. Additionally, the structure allows for external tooling to understand the type of each file, making it easier for other programs to display the downloaded assets (e.g. a simple sticker browser). Finally, filename collisions can never happen with this system while still being idempotent.

A drawback of the new system is that some of the resulting file paths are pretty ugly. If this is a major concern (which would be fairly reasonable as the file paths are essentially the implied public API of `--media`), we could bring back some amount of URL introspection to make them more pretty (e.g. moving server-specific user avatars from `guilds/{guild_id}/users/{user_id}/avatars/{avatar_hash}.png` to `avatars/{user_id}/guilds/{guild_id}/{avatar_hash}.png`.

## Further follow-ups

If this style of change is wanted as a way to fix https://github.com/Tyrrrz/DiscordChatExporter/issues/1231 _before_ v3, then I'm happy to iterate on this PR to make it production-ready. I'd imagine the following changes could be discussed before merging:
- Custom handling of twemoji links (e.g. `https://cdn.jsdelivr.net/gh/twitter/twemoji@latest/assets/svg/1f44d.svg`) to shorten the output file path (for example, to put them in a `twemojis/` folder)
- Custom handling of `images-ext-1.discordapp.net` links to shorten the output file path to just the external URL
- Better sanitation of external media URLs
  - I didn't see any invalid file paths in my testing, but I wonder if it could happen

If this isn't the direction that DCE wants to go in, then I'm happy to keep it in my personal fork. Thanks!

### DIFF
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -97,16 +97,16 @@ public abstract class ExportCommandBase : DiscordCommandBase
 
     [CommandOption(
         "reuse-media",
-        Description = "Reuse previously downloaded assets to avoid redundant requests."
+        Description = "Reuse previously downloaded assets to avoid redundant requests. "
+            + "Keep --media-nested consistent across runs for best results."
     )]
     public bool ShouldReuseAssets { get; init; } = false;
 
     [CommandOption(
-        "new-media-paths",
-        Description = "Saves media with a new file naming convention to prevent collisions. "
-            + "Duplicate media may be downloaded if consecutive runs use different values for this flag."
+        "media-nested",
+        Description = "Saves assets with a nested file naming convention, creating subdirectories for each distinct asset type."
     )]
-    public bool ShouldUseNewMediaFilePaths { get; init; } = false;
+    public bool ShouldUseNestedMediaFilePaths { get; init; } = false;
 
     [CommandOption(
         "media-dir",
@@ -162,9 +162,9 @@ public abstract class ExportCommandBase : DiscordCommandBase
         }
 
         // New media file path can only be enabled if the download assets option is set
-        if (ShouldUseNewMediaFilePaths && !ShouldDownloadAssets)
+        if (ShouldUseNestedMediaFilePaths && !ShouldDownloadAssets)
         {
-            throw new CommandException("Option --new-media-paths cannot be used without --media.");
+            throw new CommandException("Option --media-nested cannot be used without --media.");
         }
 
         // Assets directory can only be specified if the download assets option is set
@@ -283,7 +283,7 @@ public abstract class ExportCommandBase : DiscordCommandBase
                                         ShouldFormatMarkdown,
                                         ShouldDownloadAssets,
                                         ShouldReuseAssets,
-                                        ShouldUseNewMediaFilePaths,
+                                        ShouldUseNestedMediaFilePaths,
                                         Locale,
                                         IsUtcNormalizationEnabled
                                     );
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -102,6 +102,13 @@ public abstract class ExportCommandBase : DiscordCommandBase
     public bool ShouldReuseAssets { get; init; } = false;
 
     [CommandOption(
+        "new-media-paths",
+        Description = "Saves media with a new file naming convention to prevent collisions. "
+            + "Duplicate media may be downloaded if consecutive runs use different values for this flag."
+    )]
+    public bool ShouldUseNewMediaFilePaths { get; init; } = false;
+
+    [CommandOption(
         "media-dir",
         Description = "Download assets to this directory. "
             + "If not specified, the asset directory path will be derived from the output path."
@@ -152,6 +159,12 @@ public abstract class ExportCommandBase : DiscordCommandBase
         if (ShouldReuseAssets && !ShouldDownloadAssets)
         {
             throw new CommandException("Option --reuse-media cannot be used without --media.");
+        }
+
+        // New media file path can only be enabled if the download assets option is set
+        if (ShouldUseNewMediaFilePaths && !ShouldDownloadAssets)
+        {
+            throw new CommandException("Option --new-media-paths cannot be used without --media.");
         }
 
         // Assets directory can only be specified if the download assets option is set
@@ -270,6 +283,7 @@ public abstract class ExportCommandBase : DiscordCommandBase
                                         ShouldFormatMarkdown,
                                         ShouldDownloadAssets,
                                         ShouldReuseAssets,
+                                        ShouldUseNewMediaFilePaths,
                                         Locale,
                                         IsUtcNormalizationEnabled
                                     );
--- a/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
@@ -67,7 +68,7 @@ internal partial class ExportAssetDownloader
 {
     private static string GetFilePathFromUrl(string url)
     {
-        var uri = new Uri(url);
+        var uri = new Uri(NormalizeUrl(url));
 
         // Try to extract the file name from URL
         var pathAndFileName = Regex.Match(uri.AbsolutePath, @"/(.+)/([^?]*)");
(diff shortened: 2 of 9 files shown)


DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -54,6 +63,37 @@ await Http.ResiliencePipeline.ExecuteAsync(
...
+        // If this isn't a Discord CDN URL, save the file to the `media/external` folder.
+        if (!string.Equals(uri.Host, "cdn.discordapp.com", StringComparison.OrdinalIgnoreCase))
+        {
+            File.AppendAllTextAsync("external_urls.txt", $"{url}\n");
+            return $"external/{uri.Host}{uri.AbsolutePath}";
+        }
+
+        // If it is a Discord URL, we're guaranteed to have matches for these groups. <see cref="ImageCdn"/>
+        if (string.IsNullOrWhiteSpace(path) || string.IsNullOrWhiteSpace(fileName))
+            throw new Exception("Invalid Discord URL shape");
```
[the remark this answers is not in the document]
Tyrrrz: I think we should fail more gracefully here instead of an exception. I.e. treat this URL as non-Discord CDN

DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -54,6 +63,37 @@ await Http.ResiliencePipeline.ExecuteAsync(
...
+
+        // Try to extract the file name from URL
+        var pathAndFileName = Regex.Match(uri.AbsolutePath, @"/(.+)/([^?]*)");
+        var path = pathAndFileName.Groups[1].Value;
+        var fileName = pathAndFileName.Groups[2].Value;
+
+        // If this isn't a Discord CDN URL, save the file to the `media/external` folder.
+        if (!string.Equals(uri.Host, "cdn.discordapp.com", StringComparison.OrdinalIgnoreCase))
+        {
+            File.AppendAllTextAsync("external_urls.txt", $"{url}\n");
```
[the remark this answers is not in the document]
benrucker: Ah oops, this was for debugging while I was writing this and is not needed! Meant to remove it.

DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -54,6 +63,37 @@ await Http.ResiliencePipeline.ExecuteAsync(
...
+        }
+
+        // If it is a Discord URL, we're guaranteed to have matches for these groups. <see cref="ImageCdn"/>
+        if (string.IsNullOrWhiteSpace(path) || string.IsNullOrWhiteSpace(fileName))
+            throw new Exception("Invalid Discord URL shape");
+
+        // If there is a size parameter, add it as the final folder in the path.
+        // This prevents multiple sizes of an avatar, for example, from overwriting each other.
+        var sizeParam = HttpUtility.ParseQueryString(uri.Query)["size"];
+        var sizePathSegment = sizeParam != null ? $"{sizeParam}px/" : "";
```
Tyrrrz: I think we should make this more generic and instead include all query parameters into the filename (instead of making new directories). Something like:

```
avatar_size=256px.png
```

DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -54,6 +63,37 @@ await Http.ResiliencePipeline.ExecuteAsync(
...
+
+        return $"{path}/{sizePathSegment}{Path.EscapeFileName(fileName)}";
+    }
```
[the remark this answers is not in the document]
Tyrrrz: I think it's enough to duplicate existing tests and make them leverage the new option

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -101,6 +101,13 @@ public string OutputPath
     )]
     public bool ShouldReuseAssets { get; init; } = false;
 
+    [CommandOption(
+        "new-media-paths",
```
Tyrrrz: Let's use `media-nested`:

1. Avoids the "new" as that doesn't really convey anything useful to the user.
2. Explains what it actually does.

Please update related symbol names as well.

DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
```diff
@@ -101,6 +101,13 @@ public string OutputPath
...
+        Description = "Saves media with a new file naming convention to prevent collisions. "
+            + "Duplicate media may be downloaded if consecutive runs use different values for this flag."
```
Tyrrrz: I think the second part is a bit confusing in this case. We can instead update the description of `--reuse-media` to mention that the `--media-nested` should be consistent.

```suggestion
        Description = "Saves assets with a nested file naming convention, creating subdirectories for each distinct asset type."
```

DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
benrucker: Adding the tests revealed that when outputting HTML, the `=` will get URL-encoded to `%3D`, which could cause problems. So maybe instead of equals, we could use a hyphen instead? 

Ex:
`avatar_size-256_spoiler-false_animated-true.png`

benrucker:
Regarding how best to handle external URLs, I'm thinking of the following:

- [Non-controversial] Escape/sanitize each segment of the external URL to make sure it's safe for the filesystem
   - Example: https://github.com/benrucker/DiscordChatExporter/commit/eab866fe1fb402150995caf0b257d3a987ae28f5#diff-08d58480711112611abe0ffcdfc5b459a40848d2298a6bc5b18283c18f9f83fdR147-R161 
- [Only if there's buy-in] Special case `*.discordapp.net/external/` and `twemoji` from `cdn.jsdelivr.net` because these seemed exceedingly common in my (very small) testing and are probably pretty stable
  - Example: https://github.com/benrucker/DiscordChatExporter/commit/eab866fe1fb402150995caf0b257d3a987ae28f5#diff-08d58480711112611abe0ffcdfc5b459a40848d2298a6bc5b18283c18f9f83fdR106-R145

I'm less sure that the second point needs to be upstreamed, just something to consider if wanted!

DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
Tyrrrz: Yeah it's fine

Tyrrrz:
Let's not special case anything for now. File paths may become a concern later on though.

DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -44,6 +51,9 @@ await Http.ResiliencePipeline.ExecuteAsync(
             {
                 // Download the file
                 using var response = await Http.Client.GetAsync(url, innerCancellationToken);
+                var directory = Path.GetDirectoryName(filePath);
+                if (directory != null)
+                    Directory.CreateDirectory(directory);
                 await using var output = File.Create(filePath);
```
HeroponRikiBestest, line 57: At the very minimum, player.twitch.tv (likely at least some others?) embeds crash here, as the downloaded url will be something like 
```html
https://player.twitch.tv/?video=[id]&player=twitter&autoplay=true&parent=meta.tag
```
Due to the fact that the specific video is only specified via query strings on the base site url, there's no url-specified filename, and this means that the filePath variable will be set to `[media directory]/external/player.twitch.tv`, which is an already-existing directory by this point, and the program will crash trying to run File.Create on this path.

benrucker:
Thanks for the discussion on this PR! Glad something similar got merged. Thanks for the hard work

PR #1532, opened 2026-05-14
Render polls in HTML exports

adminlip:
Adds basic poll support to HTML exports by preserving the raw poll payload from Discord messages and rendering the poll question and available answers alongside the existing message content. Poll-only messages are no longer treated as empty, so they are included in the exported output instead of only showing their reaction data. Verified with `git diff --check` and targeted content assertions; a full build could not be run because `dotnet` is not installed in this environment.


Tyrrrz:
How does this compare to your other PR, #1533?

PR #1534, opened 2026-05-14
Render poll data in HTML exports

mzl2233:
This adds basic HTML rendering for Discord poll payloads by preserving the raw poll data on parsed messages, treating poll-only messages as non-empty, and outputting the poll question and answers in the message body. I could not run the .NET test suite in this environment because the `dotnet` CLI is not installed, but `git diff --check` passes.


Tyrrrz:
Superseded by #1581

PR #1536, opened 2026-05-17
Add message deletion feature (CLI & GUI)

primetime43:
## Summary
- Adds a `deletemessages` CLI command that deletes the authenticated user's own messages from a channel, with optional `--before` / `--after` date filters.
- Adds a Delete action to the GUI dashboard: a new red Delete button next to Export, a `DeleteSetup` dialog mirroring the export setup, and per-channel parallel deletion with progress reporting.
- Extends `DiscordClient` with `GetCurrentUserAsync` (so the caller's user ID can be resolved) and `DeleteMessageAsync`, plus a private `DeleteResponseAsync` helper that goes through the same resilience pipeline and respects advisory rate limits the same way the existing `GetResponseAsync` does.
- `DeleteMessageAsync` returns `true` on `204 No Content`, `false` on `403 Forbidden` / `404 Not Found` (so already-deleted or not-owned messages are skipped rather than crashing the run), and throws on anything else.

## Notes
- Only the authenticated user's messages are deleted; messages from other authors are filtered out client-side before any DELETE request is issued.
- GUI deletion uses the same `ParallelLimit` setting and progress muxer pattern as the export flow.

## Test plan
- [ ] Run `deletemessages -c <channel-id>` against a test channel and confirm only your own messages are removed.
- [ ] Run `deletemessages -c <channel-id> --after <date> --before <date>` and confirm the date range is honored.
- [ ] In the GUI, select one or more channels, click Delete, confirm the dialog, and verify per-channel progress + final summary.
- [ ] Verify rate-limiting: trigger a high-volume deletion and confirm no `429`s leak out (advisory limits respected).
- [ ] Confirm a `403` (e.g. another user's message in a guild channel) is skipped silently rather than aborting the run.

### DIFF
--- a/DiscordChatExporter.Cli/Commands/DeleteMessagesCommand.cs
+++ b/DiscordChatExporter.Cli/Commands/DeleteMessagesCommand.cs
@@ -1,0 +1,175 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CliFx.Attributes;
+using CliFx.Infrastructure;
+using DiscordChatExporter.Cli.Commands.Base;
+using DiscordChatExporter.Core.Discord;
+using DiscordChatExporter.Core.Utils.Extensions;
+
+namespace DiscordChatExporter.Cli.Commands;
+
+[Command("deletemessages", Description = "Deletes messages from a channel.")]
+public class DeleteMessagesCommand : DiscordCommandBase
+{
+    [CommandOption(
+        "channel",
+        'c',
+        Description = "Channel ID. Note: You can only delete your own messages in DMs."
+    )]
+    public required Snowflake ChannelId { get; init; }
+
+    [CommandOption(
+        "before",
+        Description = "Limit to messages sent before this date (formatted using the current culture)."
+    )]
+    public DateTimeOffset? Before { get; init; }
+
+    [CommandOption(
+        "after",
+        Description = "Limit to messages sent after this date (formatted using the current culture)."
+    )]
+    public DateTimeOffset? After { get; init; }
+
+    public override async ValueTask ExecuteAsync(IConsole console)
+    {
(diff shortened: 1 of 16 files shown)


primetime43:
@Tyrrrz can we get this reviewed and merged?

Tyrrrz:
> @Tyrrrz can we get this reviewed and merged?

It's on my todo list

CanePlayz:
This is an interesting use case of DCE

primetime43:
@Tyrrrz are you going to complete this? Fix the conflict or me?

SnowySailor:
This doesn't seem like it belongs in this application, just my opinion. Doing a quick google search there are plenty of ways that you can purge all of your messages. Here are some examples:

https://github.com/victornpb/undiscord
https://redact.dev/services/discord
https://chromewebstore.google.com/detail/discrub-discord-message-m/plhdclenpaecffbcefjmpkkbdpkmhhbj?pli=1

Is there a reason to not use one of these? This app is an exporter, and I don't think a destructive message deletion feature belongs here.

primetime43:
> This doesn't seem like it belongs in this application, just my opinion. Doing a quick google search there are plenty of ways that you can purge all of your messages. Here are some examples:
> 
> https://github.com/victornpb/undiscord https://redact.dev/services/discord https://chromewebstore.google.com/detail/discrub-discord-message-m/plhdclenpaecffbcefjmpkkbdpkmhhbj?pli=1
> 
> Is there a reason to not use one of these? This app is an exporter, and I don't think a destructive message deletion feature belongs here.

Thanks, you may be right on this. I thought it was a nice feature to have, but it may be out of scope of this project's intentions like you said.

Tyrrrz:
After some deliberation, I agree, I don't see message deletion as being in scope for this project.

PR #1552, opened 2026-06-15
Catch both old uppercase 5-char hash and newer lowercase 5-char hash when migrating to new hash

### LINKED ISSUE #1551: New file hash generator uses lowercase while old file hash generator used uppercase, causing files to get missed during conversion
### Version

v2.47.2

### Flavor

CLI (Command-Line Interface)

### Platform

Docker / Linux

### Export format

HTML

### Steps to reproduce

1. Have posted an embedded link to some media via Twitter or some other service
2. Have that media exported via this discord chat exporter on a version older than 2.47 (when the new hash length was implemented), so the media gets exported using the old hash length of 5
3. The media is deleted from Twitter/wherever so the original source is no longer available
4. Update to 2.47 or later
5. Re-export with reuse media = true. The new export no longer contains the link to the media that was already downloaded, and the exporter attempts to re-fetch the media from the source but since it's no longer available (step 3) it hits a 404 and does not properly generate an export in html. The html links to the original discord proxy url instead of to the local path that contains the media.

### Details

I believe the issue is actually pretty simple:
The file name (with the hash) that was generated using the old filename generation function generates something like this:
`Gz8OIAQ_1-FDF13.jpg`
However, the new legacy filename generation function (`GetLegacyFileNameFromUrl`) generates this:
`Gz8OIAQ_1-fdf13.jpg`

Notice how the hash is now in **lower case**. In older versions, the hash generated was NOT converted to lower-case:
https://github.com/Tyrrrz/DiscordChatExporter/blob/97cb8e2347e07743715cf10889cf2655ac985ed9/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs#L78

This switch from uppercase to lowercase hashes was introduced in 2.46.1: https://github.com/Tyrrrz/DiscordChatExporter/compare/2.46...2.46.1#diff-08d58480711112611abe0ffcdfc5b459a40848d2298a6bc5b18283c18f9f83fdR78

With this commit: https://github.com/Tyrrrz/DiscordChatExporter/commit/768e124370c42c62c3eef269b8d039dbe6beeed6

This means that during the check for existing media, we are looking for the wrong file in some cases:
https://github.com/Tyrrrz/DiscordChatExporter/pull/1485/changes#diff-08d58480711112611abe0ffcdfc5b459a40848d2298a6bc5b18283c18f9f83fdR42-R125

The fix here is to check for not only lower-case 5-character hashes but to also check for uppercase hashes. Since it's a quick operation, it should not impact performance.

### Checklist

SnowySailor:
This fixes a bug with the media export

Originally, file hashes were generated using a 5-character all **uppercase** hex hash. In a more recent version of this tool (2.46.1, [this commit](https://github.com/Tyrrrz/DiscordChatExporter/commit/768e124370c42c62c3eef269b8d039dbe6beeed6)), it was changed to a 5-character all **lowercase** hex hash. In the most recent version, the 5-character hex hash has been updated to a 16-character hash, with support for renaming files that were generated using the previous hash method.

However, this renaming process does not include a check for the original 5-character uppercase hash file. So files that were generated with the uppercase hash are ignored, and only files with the lowercase hash are detected and renamed. If a piece of media was deleted at the source (e.g. deleted from twitter) and can no longer be re-fetched, that media is treated as inaccessible even if the media was already exported using the 5-char uppercase file hash. **The file exists on disk already from a previous export, but the code does not detect it since it's only looking for files with a lowercase hash.**

This PR adds support for detecting files that were generated with the original uppercase hash as well as the newer lowercase hash.

Closes #1551

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
@@ -149,11 +149,7 @@ internal partial class ExportAssetDownloader
         return
         [
             // Lowercase variant (introduced in 2.46.1)
-            GetFileNameFromUrl(
-                url,
-                Convert.ToHexStringLower(hashData).Truncate(5)
-            ),
-
+            GetFileNameFromUrl(url, Convert.ToHexStringLower(hashData).Truncate(5)),
             // Uppercase variant (original)
             GetFileNameFromUrl(
                 url,
--- a/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
@@ -38,8 +38,8 @@ internal partial class ExportAssetDownloader(string workingDirPath, bool reuse)
             return _previousPathsByUrl[url] = filePath;
 
         // Check for a file cached by the legacy naming scheme (5-char hash) and rename it
-        // to the new naming scheme to preserve backwards compatibility with existing exports
-        // This will catch both the 5-char lowercase hash and the 5-char uppercase hash variants
+        // to the new naming scheme to preserve backwards compatibility with existing exports.
+        // This will catch both the 5-char lowercase hash and the 5-char uppercase hash variants.
         if (reuse)
         {
             var legacyFileNames = GetLegacyFileNamesFromUrl(url);
@@ -52,7 +52,7 @@ internal partial class ExportAssetDownloader(string workingDirPath, bool reuse)
                     // earlier existence check and this move operation
                     try
                     {
-                        File.Move(legacyFilePath, filePath, overwrite: true);
+                        File.Move(legacyFilePath, filePath, true);
                         return _previousPathsByUrl[url] = filePath;
                     }
                     catch (IOException)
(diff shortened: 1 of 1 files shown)


DiscordChatExporter.Core/Exporting/ExportAssetDownloader.cs
```diff
@@ -137,13 +142,23 @@ private static string GetFileNameFromUrl(string url) =>
         );
 
     // Legacy naming used a 5-char hash, kept for backwards compatibility with existing exports
-    private static string GetLegacyFileNameFromUrl(string url) =>
-        GetFileNameFromUrl(
-            url,
-            SHA256
-                .HashData(Encoding.UTF8.GetBytes(NormalizeUrl(url)))
-                .Pipe(Convert.ToHexStringLower)
```
Tyrrrz, line 145: Wouldn't a simpler fox be to just use `ToHexString` here instead of `ToHexStringLower`?

SnowySailor: There was a period of time where the 5-character ToHexStringLower legacy hash was in use in a release build, so we want to make sure we maintain support of that version of the hash as well. If we just change it to ToHexString then that will break the file migration for users that used a release build of the exporter that used ToHexStringLower. Doing it this way ensures we maintain support for both previous versions.

Tyrrrz (approved):

PR #1555, opened 2026-06-16
Validate output path early to prevent errors when exporting multiple channels

### LINKED ISSUE #1549: --include-threads checks for errors only after fetching threads, which is frustrating
### Version

v2.47.2

### Flavor

CLI (Command-Line Interface)

### Platform

Windows

### Export format

_No response_

### Steps to reproduce

Export a server with --include-threads and do something that will give an error

### Details

Fetching channels...
Fetched 54 channel(s).
Fetching threads...
Fetched 334 thread(s).
Attempted to export multiple channels, but the output path is neither a directory nor a template. If the provided output path is meant to be treated as a directory, make sure it ends with a slash. Provided output path: ...

It shouldn't check for errors only after fetching threads, which takes a long time. It should check for any errors before fetching anything. Currently it is frustrating having it to fetch threads and get an error, and having to fetch threads again to try to fix it.

### Checklist

CanePlayz:
Closes #1549

This PR fixes #1549 by moving the output path validation before thread fetching starts.

### DIFF
--- a/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
+++ b/DiscordChatExporter.Cli/Commands/Base/ExportCommandBase.cs
@@ -167,6 +167,37 @@ public abstract class ExportCommandBase : DiscordCommandBase
 
         var unwrappedChannels = new List<Channel>(channels);
 
+        // Make sure the user does not try to export multiple channels into one file.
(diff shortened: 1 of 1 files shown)


Tyrrrz (approved):

Tyrrrz:
Very elegant, thanks!

PR #1569, opened 2026-07-13
Fix reverse exports with an after boundary

ShiroKSH:
## Summary

- Stop reverse pagination at the requested after boundary.
- Add DateRange coverage for reverse exports with an after boundary.

## Root cause

Reverse pagination used the after boundary to calculate progress but kept fetching and exporting older pages.

## Validation

- Release build passed.
- Formatting verification passed.

### DIFF
--- a/DiscordChatExporter.Core/Discord/DiscordClient.cs
+++ b/DiscordChatExporter.Core/Discord/DiscordClient.cs
@@ -796,6 +796,10 @@ public class DiscordClient(
 
             foreach (var message in messages)
             {
+                // Messages are ordered newest to oldest, so no later message can satisfy this bound.
+                if (after is not null && message.Id.CompareTo(after.Value) <= 0)
+                    yield break;
+
                 lastMessage ??= message;
 
                 // Report progress based on timestamps
--- a/DiscordChatExporter.Core/Discord/DiscordClient.cs
+++ b/DiscordChatExporter.Core/Discord/DiscordClient.cs
@@ -796,8 +796,8 @@ public class DiscordClient(
 
             foreach (var message in messages)
             {
-                // Messages are ordered newest to oldest, so no later message can satisfy this bound.
-                if (after is not null && message.Id.CompareTo(after.Value) <= 0)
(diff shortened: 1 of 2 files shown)


DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -796,6 +796,10 @@ public async IAsyncEnumerable<Message> GetMessagesInReverseAsync(
 
             foreach (var message in messages)
             {
+                // Messages are ordered newest to oldest, so no later message can satisfy this bound.
+                if (after is not null && message.Id.CompareTo(after.Value) <= 0)
+                    yield break;
```
Tyrrrz: The non-reverse export makes a similar check like so (lines 715-717):

```csharp
// Ensure that the messages are in range
if (message.Timestamp > lastMessage.Timestamp)
    yield break;
```

Can we adapt the check you added so it's symmetrically consistent with this one?

Example: `message.Timestamp <= after?.ToDate()` (make sure the logic is equivalent)

ShiroKSH: done

DiscordChatExporter.Core/Discord/DiscordClient.cs
```diff
@@ -796,6 +796,10 @@ public async IAsyncEnumerable<Message> GetMessagesInReverseAsync(
 
             foreach (var message in messages)
             {
+                // Ensure that the messages are in range
+                if (message.Timestamp <= after?.ToDate())
```
Tyrrrz, line 800:
```suggestion
                if (message.Timestamp <= firstMessage.Timestamp)
```

Would that not also be correct? 🤔 
Or is there a specific reason we can't do that here?

ShiroKSH: 4 my mind no firstMessage is the newest message after the boundary, so this would stop on the first page. We need to compare against after; ideally by Snowflake ID to avoid losing same-millisecond precision.

PR #1576, opened 2026-07-28
Only allow http(s) URLs in exported HTML links (fixes #1575)

### LINKED ISSUE #1575: Stored XSS in HTML exports through embed URLs (#1568 fixed only the markdown path)
### Version

`prime` @ `0358045` (the commit that merged #1568)

### Export format

HTML

### Summary

#1568 fixed the masked-link XSS from #1567 by restricting the markdown link matchers to `https?://`. Embed URLs reach the same anchor sink but do not go through the markdown parser, so they were not covered. A `javascript:` URL in an embed still renders as a live `href` in HTML exports.

### Affected sinks

`DiscordChatExporter.Core/Exporting/MessageGroupTemplate.cshtml`:

| line | sink |
|---|---|
| 489 | `<a class="chatlog__embed-author-link" href="@embed.Author.Url">` |
| 507 | `<a class="chatlog__embed-title-link" href="@embed.Url">` |
| 609 | `<a class="chatlog__embed-author-link" href="@embed.Author.Url">` |
| 627 | `<a class="chatlog__embed-title-link" href="@embed.Url">` |

`Embed.Parse` and `EmbedAuthor.Parse` read `url` with `GetNonWhiteSpaceStringOrNull()` and apply no scheme check; the template guards only on `!string.IsNullOrWhiteSpace(...)`. Razor encodes the value, which protects the attribute syntax but does nothing about the scheme — the same point #1567 made about the markdown path. These two fields are links rather than media, so unlike the embed image/icon fields there is no `proxy_url` fallback that would sanitise them.

### Reproduction

`embed.url` and `embed.author.url` are only settable via a webhook or bot, not by a normal message, so the payload is delivered through a webhook posted to the channel:

```json
{
  "embeds": [{
    "type": "rich",
    "title": "Click me (embed title link)",
    "url": "javascript:document.title='DCE-EMBED-TITLE-XSS'",
    "author": {
      "name": "Click me (embed author link)",
      "url": "javascript:document.title='DCE-EMBED-AUTHOR-XSS'"
    }
  }]
}
```

Export the channel to HTML, then grep the output:

```
href="javascript:document.title='DCE-EMBED-AUTHOR-XSS'"
href="javascript:document.title='DCE-EMBED-TITLE-XSS'"
```

Clicking either link in the exported page runs the script. I verified this against `prime` at `0358045` by building the CLI and serving the payload above from a local stand-in for the Discord API, so the export runs through the normal `ChannelExporter` path.

### Threat model

Weaker than #1567, and worth stating plainly:

- #1567 needed only the ability to post a message. This needs the ability to set an embed `url`, i.e. a webhook or bot in the channel being exported. That is still common, since Manage Webhooks is widely delegated and forwarded messages carry embeds in from elsewhere.
- It requires the victim to click the crafted link in the export, whereas #1567's masked link was also a click. It does not execute during the export itself.

The person at risk is someone archiving a channel they do not control, which is a normal reason to run DCE. When the export is opened from disk the script runs in the `file://` origin (read/exfiltrate the archived conversation, rewrite what the page shows). When the export is published to a web host — a common thing to do with these logs — it becomes stored XSS on that origin.

### Suggestion

The parser is the wrong layer to fix this, which is why #1568 did not generalise: a regex was carrying the security property. Every `href` written to the export comes from a small number of sinks, so a single helper would cover the embed cases and anything added later:

```csharp
private static string? SafeUrl(string? url) =>
    Uri.TryCreate(url, UriKind.Absolute, out var uri)
    && (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps)
        ? url
        : null;
```

Apply it at each `href`/`src` sink and drop the anchor when it returns null. `HtmlMarkdownVisitor.VisitLinkAsync` could then rely on the same helper rather than on the three regexes.

Happy to open a PR if you'd like it done this way.

---

Unrelated hardening note while in this area: with `--media`, `ExportAssetDownloader.DownloadAsync` fetches embed image/icon URLs directly. Fetching third-party hosts is inherent to archiving embeds, so that is expected, but the downloader does not refuse loopback or private ranges (`127.0.0.1`, `169.254.169.254`, `192.168.x`), which can never be a legitimate embed image. Blocking those would be a cheap SSRF guard. Happy to fold it into the same PR or leave it out your call.

MildlyMeticulous:
Fixes #1575.

`EnsureSafeUrl` on `ExportContext`, as you suggested, returning the URL only when it parses as an absolute http or https URI:

```csharp
public static string? EnsureSafeUrl(string? url) =>
    Uri.TryCreate(url, UriKind.Absolute, out var uri)
    && (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps)
        ? url
        : null;
```

Applied at the four embed anchors. Each already had an else branch rendering the title or author as a plain div, so an unsafe URL now takes that path and the text still shows, just unlinked.

`HtmlMarkdownVisitor.VisitLinkAsync` uses the same helper and renders the link children without an anchor when the scheme is not allowed. That way the three regexes in `MarkdownParser` are no longer the only thing keeping unsafe schemes out of the export, which is what let #1568 miss the embed path.

Verified against `prime` by serving the payload from a local stand-in for the Discord API:

- `javascript:` and `data:` in `embed.url` and `embed.author.url` produce no `href`, and the title and author text still render
- `https://example.com/page?a=1&b=2` and `http://example.org/author` still render as links

The `Cli.Tests` specs export live channels, so covering this in the suite needs a fixture channel carrying an embed with an author URL and a title URL. Point me at one and I'll add the specs.

Left the private-range question out, as discussed. It needs a `ConnectCallback` rather than a scheme check, so it does not belong in this change.

### DIFF
--- a/DiscordChatExporter.Core/Exporting/ExportContext.cs
+++ b/DiscordChatExporter.Core/Exporting/ExportContext.cs
@@ -120,6 +120,12 @@ internal class ExportContext(DiscordClient discord, ExportRequest request)
     public Color? TryGetUserColor(Snowflake id) =>
         GetUserRoles(id).Where(r => r.Color is not null).Select(r => r.Color).FirstOrDefault();
 
+    public static string? EnsureSafeUrl(string? url) =>
+        Uri.TryCreate(url, UriKind.Absolute, out var uri)
+        && (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps)
+            ? url
+            : null;
+
     public async ValueTask<string> ResolveAssetUrlAsync(
         string url,
         CancellationToken cancellationToken = default
--- a/DiscordChatExporter.Core/Exporting/HtmlMarkdownVisitor.cs
+++ b/DiscordChatExporter.Core/Exporting/HtmlMarkdownVisitor.cs
@@ -188,12 +188,19 @@ internal partial class HtmlMarkdownVisitor(
             .Groups[1]
             .Value;
 
+        var safeUrl = ExportContext.EnsureSafeUrl(link.Url);
+        if (string.IsNullOrWhiteSpace(safeUrl))
+        {
+            await VisitAsync(link.Children, cancellationToken);
+            return;
+        }
+
         buffer.Append(
             !string.IsNullOrWhiteSpace(linkedMessageId)
                 // lang=html
-                ? $"""<a href="{HtmlEncode(link.Url)}" onclick="scrollToMessage(event, '{linkedMessageId}')">"""
+                ? $"""<a href="{HtmlEncode(safeUrl)}" onclick="scrollToMessage(event, '{linkedMessageId}')">"""
                 // lang=html
-                : $"""<a href="{HtmlEncode(link.Url)}">"""
+                : $"""<a href="{HtmlEncode(safeUrl)}">"""
         );
 
         await VisitAsync(link.Children, cancellationToken);
--- a/DiscordChatExporter.Core/Exporting/MessageGroupTemplate.cshtml
+++ b/DiscordChatExporter.Core/Exporting/MessageGroupTemplate.cshtml
@@ -24,6 +24,9 @@
     string FormatDate(DateTimeOffset instant, string format = "g") =>
         Context.FormatDate(instant, format);
 
+    string? EnsureSafeUrl(string? url) =>
+        ExportContext.EnsureSafeUrl(url);
+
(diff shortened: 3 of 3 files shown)


Tyrrrz:
The test message created here: https://discordapp.com/channels/866458392705105940/866472452459462687/1533115193692000257

image

Tyrrrz:
@MildlyMeticulous are you still planning to work on this PR?

PR #1589, opened 2026-09-09
--threads-output parameter and new template tokens to aid in thread file paths

### LINKED ISSUE #1123: Make template tokens work properly for export paths of threads

Togekiss:
Closes #1123

Hi! This is my own take on the years old issue of solving paths template tokens to export threads.

I'm aware of other discussions, and the possibility of it being deprecated by a CLI arguments rework... but I had to make it to simplify my own workflow, so I might as well share my approach!

The philosophy is to not modify existing mechanics by adding an optional parameter :)

## New parameter for 'export' commands: --threads-output

If you don't use the parameter, or use only `-o`, paths will resolve in the exact same way as they always do.
Using the parameter allows you to configure thread paths without worrying about how it affects regular channel paths!

Example: `DCE export -t "mfa.Ifrn" -c 53555 -o "C:\Export\%G\%T\%C.html" --threads-output "C:\Export\%G\Threads\%T#  %C.html" --include-threads all`
- Channel is saved to `C:\Export\Server\Category\my-channel.html`
- Thread is saved to:
  -  Without `--threads-output`: `C:\Export\Server\my-channel\My Thread.html`
  -  With `--threads-output`: `C:\Export\Server\Threads\my-channel# My Thread.html`

## New template tokens: %m, %M, %N

Existing template tokens aren't changed, they always resolve to the target's direct parent (category for channels, channel for threads):
- `%t` - parent ID
- `%T` - parent name
- `%P` - parent position

The new template tokens **always** resolve to the containing category.
- `%m` - category ID
- `%M` - category name
- `%N` - category position

This allows to do something like: `DCE export -t "mfa.Ifrn" -c 53555 -o "C:\Export\%G\%M\%C.html" --include-threads all`
- Channel is saved to `C:\Export\Server\Category\my-channel.html`
- Thread is saved to `C:\Export\Server\Category\My Thread.html`

## Combining them

Now we can configure a different path logic for threads, and have the flexibility to combine category, channel and thread template tokens however we want.

Example: `DCE export -t "mfa.Ifrn" -c 53555 -o "C:\Export\%G\%T\%C.html" --threads-output "C:\Export\%G\%M\%T threads\%C.html" --include-threads all`
- Channel is saved to `C:\Export\Server\Category\my-channel.html`
- Thread is saved to `C:\Export\Server\Category\my-channel threads\My Thread.html`

## Disclaimer

At the moment it's a CLI-exclusive feature and I haven't tested it against forums. It works perfectly as it is for me, but if you think this is valuable we could look more into it!

### DIFF
--- a/.docs/Using-the-CLI.md
+++ b/.docs/Using-the-CLI.md
@@ -111,26 +111,53 @@ You can use template tokens to generate the output file path based on the server
 ./DiscordChatExporter.Cli export -t "mfa.Ifrn" -c 53555 -o "C:\Discord Exports\%G\%T\%C.html"
 ```
 
-Assuming you are exporting a channel named `"my-channel"` in the `"Text channels"` category from a server
-called `"My server"`, you will get the following output file
-path: `C:\Discord Exports\My server\Text channels\my-channel.html`
+Assuming you are exporting a channel named `"my-channel"` in the `"Text channels"` category from a server called `"My server"`, you will get the following output file path: `C:\Discord Exports\My server\Text channels\my-channel.html`
 
 Here is the full list of supported template tokens:
 
 - `%g` - server ID
 - `%G` - server name
-- `%t` - category ID
-- `%T` - category name
+- `%t` or `%m` - category ID
+- `%T` or `%M` - category name
 - `%c` - channel ID
 - `%C` - channel name
 - `%p` - channel position
-- `%P` - category position
+- `%P` or `%N` - category position
 - `%a` - the "after" date
 - `%b` - the "before" date
 - `%d` - the current date
 - `%%` - escapes `%`
 
-#### Partitioning
+#### Setting a different output directory and/or filenames for threads
+
+By default, threads will follow the same output path as all channels, using either the default or the path set with `-o`.
+
+But note that if the channel being imported is a thread, some template tokens will resolve differently:
+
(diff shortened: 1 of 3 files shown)


Tyrrrz:
Note that this PR doesn't fully close #1123 since it only implements this functionality for the CLI

Togekiss:
I'll look into adding it to the GUI if you think it's worthwhile! I fixed the documentation mishaps Copilot found.

(I see some supressed review comments too, but I figure there's nothing to do there if they're supressed?)

Togekiss:
Putting it in the GUI wasn't too bad! Everything works on my end. I had to rely on machines for German and Ukrainian translations, I hope they are okay!

